PR #8, opened 2018-09-06
Feature/implement private store

oed:
This PR implements the private store using orbitdb as a data store, and muport to do encryption.


src/privateStore.js
```diff
@@ -65,30 +72,52 @@ class PrivateStore {
...
+          resolve()
+        })
+        this.db.sync(log.values)
+      })
+    }
+    // This is the first time the store is used.
+    // Generate a random salt and save in the db.
+    this.salt = Buffer.from(nacl.randomBytes(16)).toString('hex')
+    const encryptedSalt = this._encryptEntry(this.salt)
+    await this.db.put(SALT_KEY, encryptedSalt)
```
ajunge: Why do we need all this SALT_KEY thing?

src/privateStore.js
```diff
@@ -65,30 +72,52 @@ class PrivateStore {
    * @param     {String}    hash                        The hash of the private store OrbitDB
    */
   async _sync (hash) {
-    // if hash === null; create a new orbitdb instance.
-    const orbitdb = new OrbitDB(this.ipfs)
-    //TODO - sync the OrbitDB key-value store
+    if (!this.db) {
+      const orbitdb = new OrbitDB(this.ipfs)
+      this.db = await orbitdb.keyvalue('3box.datastore', {
```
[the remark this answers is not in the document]
oed: Good point, right now it doesn't matter since we only sync using hashes anyway. However if we at some point in the future enable syncing with pubsub this could screw things up. We should probably put some unique name for each user here.

c-castillo:
@oed there is a confict in your PR

oed:
@c-castillo tnx, should be up to date now.

src/privateStore.js
```diff
@@ -38,14 +43,16 @@ class PrivateStore {
...
+    if (value != null) {
+      value = this._encryptEntry(value)
+    }
     const dbKey = this._genDbKey(key)
 
     // TODO - error handling
-    const multihash = await this.db.put(dbKey, encryptedValue)
-
-    this.updateRoot(multihash)
+    const hash = await this.db.put(dbKey, value)
```
christianlundkvist: So are we setting the value "null" here if the user plugs in "null"? Is that intended?

src/privateStore.js
```diff
@@ -65,30 +72,58 @@ class PrivateStore {
    * @param     {String}    hash                        The hash of the private store OrbitDB
    */
   async _sync (hash) {
-    // if hash === null; create a new orbitdb instance.
-    const orbitdb = new OrbitDB(this.ipfs)
-    //TODO - sync the OrbitDB key-value store
+    if (!this.db) {
+      const orbitdb = new OrbitDB(this.ipfs)
+      // the db needs a unique name, we use the hash of the DID + a store specific name
+      const storeName = multihash(this.muportDID.getDid()) + ".datastore"
```
christianlundkvist: maybe give `multihash` a name that clarifies which hash function we are using?

src/privateStore.js
```diff
@@ -65,30 +72,58 @@ class PrivateStore {
...
-// some sample code for  creating a log store
-//orbitdb.log('hello', {replicate: false}).then(db => {
-  //db.add('asdfasdf').then(h => {
-    //const eve = db.get(h)
-    //console.log(eve)
-  //})
-//}).catch(console.log)
+const multihash = str => {
+  const dataBuf = Buffer.from(str, 'utf8')
+  return Multihash.encode(dataBuf, 'sha3-256').toString('hex')
```
christianlundkvist: Why the use of `sha3-256` here? Why not `sha2-256` (more standard) or `keccak256` (more ethereum-y)?

src/privateStore.js
```diff
@@ -38,14 +43,16 @@ class PrivateStore {
    */
   async set (key, value) {
     if (!this.db) throw new Error('_sync must be called before interacting with the store')
+    if (key === SALT_KEY) throw new Error('Invalid key')
```
christianlundkvist: I guess technically the key `SALT_KEY` could be allowed because the real DB key would be `Hash(salt + "SALT_KEY")`. But I agree it's less confusing to just treat it as off-limits.

c-castillo:
Are we good to merge @ajunge ?

src/privateStore.js
oed: Yes, setting to `null`is the only way to remove an entry.

src/privateStore.js
oed: Will do.

src/privateStore.js
oed: Tnx, will update

src/privateStore.js
oed: Oh, good point. I'll just remove it from here then. I guess this code could cause more confusion than it helps.

oed:
@christianlundkvist Addressed some of your comments.

src/privateStore.js
oed: Are you saying that we should encrypt `null` so that you can't see that you've removed an entry?

christianlundkvist: @oed Is this an orbit-db convention? i.e. setting a value to null is how you delete key-value entries? If that's the case then we should probably support it as you've done.

oed: @christianlundkvist No I'm not sure there is a convention. Actually let me ask

Issue #3, opened 2018-09-05
Unit tests!


ajunge:
No tests. We need to write unit test. Jest ?

oed:
Take a look at this branch: https://github.com/uport-project/3box-js/tree/feature/implement-private-store

ajunge:
Looks good. Any reason not to mockup ipfs-api and instead spawn an ipfs process?

oed:
Yeah, it's a **lot** more work. Especially with orbitdb, don't know exactly what methods it is using. 
Is there a reason to **not** spawn an ipfs process?

ajunge:
The idea of unit test is to isolate the code from outside factors. I think we can mockup ipfs orbit-db or any code that is not part of our code

oed:
I'd say that the purpose of a unit test is to test if the specific unit works as intended, regardless of how the entire system works. The purpose of the "store" units is to communicate correctly with ipfs and orbitdb. So if we mock up these modules and do some small minor error (which is very likely) we'll end up building a unit that works great with our mocks but not with the actual modules we are using.

ipfsd-cli is built for this purpose, to test code isolated from any real ipfs node (such as infura). So in a sense we are using a mock ipfs instead of the real api we use in production.

PR #21, opened 2018-09-10
Feature/implement get profile

oed:
Should be simple enough.


oed:
* Note this should not be merged before #20 !

Tnx, @c-castillo I'm still waiting for andres to finish #3 then it should be easy to add!

PR #31, opened 2018-09-13
Add documentation to close and logout methods

oed:
I just needed to add documentation to the close and logout methods.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -174,16 +174,18 @@ class ThreeBox {
   }
 
   /**
-   * Uninitializes the 3box. Should be called after you are done with
-   * the instance (unless logout has been called)
+   * Closes the 3box instance without clearing the local cache.
+   * Should be called after you are done using the 3Box instance,
+   * but without logging the user out.
    */
   async close () {
     await this.privateStore.close()
   }
 
   /**
-   * Uninitializes the 3box and clears the local cache. If you call this
-   * users will need to sign a message the next time you call openBox
+   * Closes the 3box instance and clears local cache. If you call this,
+   * users will need to sign a consent message to log in the next time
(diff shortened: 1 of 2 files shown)


src/3box.js
```diff
@@ -173,10 +173,18 @@ class ThreeBox {
     }
   }
 
+  /**
+   * Uninitializes the 3box. Should be called after you are done with
```
michaelsena: Uninitializes the user's 3Box without clearing the local cache. Should be called after you are done using the 3Box instance, but without logging the user out.

src/3box.js
```diff
@@ -173,10 +173,18 @@ class ThreeBox {
...
+   * the instance (unless logout has been called)
+   */
   async close () {
     await this.privateStore.close()
   }
 
+  /**
+   * Uninitializes the 3box and clears the local cache. If you call this
```
michaelsena: Uninitializes the user's 3Box and clears local cache. If you call this, users will need to sign a consent message to log in the next time you call openBox.

src/3box.js
c-castillo: It can be unclear what `Uninitializes the 3box` actually means.

oed: @c-castillo do you have a suggestion of what we could say instead? Basically we are just closing the orbitdb instance correctly.

c-castillo: `Closes the 3box instance` or more technical `Closes the underlying storage instance of the 3box`
Maybe the first one is better

oed: Ok, thanks!

oed:
@michaelsena @c-castillo Pushed an update!

c-castillo:
LGTM

PR #33, opened 2018-09-13
Feature/use del to remove

christianlundkvist:
Using the orbit-db command `db.del` to implement `remove`.

Also increased timeouts in the jest tests to 20s since the default 5s timeouts got triggered.

### DIFF
--- a/src/privateStore.js
+++ b/src/privateStore.js
@@ -41,7 +41,6 @@ class PrivateStore {
    * @return    {Boolean}                           true if successful
    */
   async set (key, value) {
-	
     if (!this.db) throw new Error('_sync must be called before interacting with the store')
 
     if (value == null) {
@@ -65,13 +64,12 @@ class PrivateStore {
    * @return    {Boolean}                           true if successful
    */
   async remove (key) {
-
     if (!this.db) throw new Error('_sync must be called before interacting with the store')
 
     const dbKey = this._genDbKey(key)
     const hash = await this.db.del(dbKey)
     this.updateRoot(hash)
-    
+
     return true
   }
 
--- a/src/privateStore.js
+++ b/src/privateStore.js
@@ -41,11 +41,15 @@ class PrivateStore {
    * @return    {Boolean}                           true if successful
    */
   async set (key, value) {
+	
     if (!this.db) throw new Error('_sync must be called before interacting with the store')
 
-    if (value != null) {
-      value = this._encryptEntry(value)
+    if (value == null) {
+      // Do nothing for null value
+      // To remove a value, call remove
+      return false
     }
+    value = this._encryptEntry(value)
     const dbKey = this._genDbKey(key)
 
     // TODO - error handling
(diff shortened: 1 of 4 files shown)


christianlundkvist:
Failed in linting. Will update.

src/privateStore.js
```diff
@@ -61,7 +64,13 @@ class PrivateStore {
...
    */
   async remove (key) {
-    return this.set(key, null)
+    if (!this.db) throw new Error('_sync must be called before interacting with the store')
+
+    const dbKey = this._genDbKey(key)
+    const hash = await this.db.del(dbKey)
+    this.updateRoot(hash)
+
+    return true
```
oed: Isn't strictly part of this story, but can you change the last three lines to 
```js
return this.updateRoot(hash)
```
And please do the same thing for the `set` method :)

src/privateStore.js
```diff
@@ -43,9 +43,12 @@ class PrivateStore {
   async set (key, value) {
     if (!this.db) throw new Error('_sync must be called before interacting with the store')
 
-    if (value != null) {
-      value = this._encryptEntry(value)
+    if (value == null) {
+      // Do nothing for null value
+      // To remove a value, call remove
+      return false
```
oed: Do we really need to check this? I'd just remove it for simplicity. Not really any need to stop people from setting stuff to null.

oed:
Thanks for fixing the timeout thing also! I think it was causing issues on circleci sometimes.

src/privateStore.js
christianlundkvist: Let me can leave the behavior the way it was regarding setting `null`.

src/privateStore.js
christianlundkvist: Yeah, this is nice. I'll make this change.

christianlundkvist:
@oed Thanks for the comments! Updated.

src/privateStore.js
```diff
@@ -46,12 +46,12 @@ class PrivateStore {
     if (value != null) {
       value = this._encryptEntry(value)
```
oed: I don't get it, why should we check if the value is null? Can't we just always do the encryption regardless?

christianlundkvist: So I'm not sure what encrypting null would mean? How is `null` represented as a `Uint8Array`? And when you decrypt, would you get back `null` or something else like the empty string?

oed: Ok, that makes sense. I guess then we would need to think about if value is `undefined` also

christianlundkvist: @oed Actually I verified that `(x != null)` returns `true` also if `x` is `undefined.` :smiley:

c-castillo: @christianlundkvist `if (value)` works for both `null` and `undefined`. Javascript is weird sometimes.

christianlundkvist: @oed Hmm that would mean that the value `false` would also trigger this. I’m actually wondering if we are even able to encrypt anything that’s not a string? In EIP 1024 I think they’re solving it by only encrypting/decrypting stringified JSON.

oed: Right, didn't think of that. So you are saying that we just do
```js
value = JSON.stringify(value)
```
before encrypting, and
```js
value = JSON.parse(value)
```
after decrypting?

That makes sense. We can put this in the `_encryptEntry` and `_decryptEntry` functions.

christianlundkvist: @oed Yes, I think that's a good way of solving it. Not sure how things like `undefined` are stringified, but this way we should be able to handle all the relevant cases.

christianlundkvist:
@oed Updated encryption/decryption with `JSON.serialize()` and `JSON.parse()`.

PR #56, opened 2018-09-28
Feature/custom keystore

oed:
To be merged after #48


oed:
@zachferland yes, exactly! The orbitdb team are working on a more sophisticated "identity provider" system where access control can be defined in a better way. I think we can migrate to that fairly simple once it's released :)

PR #57, opened 2018-09-28
fix: (un)pad encrypted values in private store

zachferland:


zachferland:
resolves https://github.com/uport-project/3box-js/issues/38

src/privateStore.js
```diff
@@ -46,11 +46,19 @@ class PrivateStore extends KeyValueStore {
...
+    return this.muportDID.symEncrypt(this._pad(JSON.stringify(entry)))
   }
 
   _decryptEntry ({ ciphertext, nonce }) {
-    return JSON.parse(this.muportDID.symDecrypt(ciphertext, nonce))
+    return JSON.parse(this._unpad(this.muportDID.symDecrypt(ciphertext, nonce)))
+  }
+
+  _pad (val, blockSize = 24) {
+    return `${val}${'\0'.repeat(blockSize - (val.length % blockSize))}`
```
christianlundkvist: @zachferland If the value to be padded is already a multiple of `blockSize` we still add another `blockSize` number of zeros, is that intended?

zachferland: nope, good point, no need to pad in that case, I can adjust that

zachferland:
@christianlundkvist made change for your comment, take a look again

oed:
@zachferland _pad and _unpad don't need to be instance methods. Do you think it would make more sense to just add them as separate functions?

zachferland:
yeah, but on the other hand just had find a place for them for now and this was the only place they are used. Wasn't concerned with placing them here for now since these are just private methods, if needed elsewhere they could be moved to a utils/crypto or something at that time. @oed any other place you would prefer them?

oed:
@zachferland I'd just make them standalone functions that are outside the class like so:
```js
function pad (data) {}
```
But yeah, I don't think it matters that much. Feel free to merge!

PR #60, opened 2018-09-28
fix: get hash digest, then encode as multihash

zachferland:


oed:
@zachferland can you help me understand why this is needed?

PR #61, opened 2018-09-28
Feat/sig ent key gen

zachferland:
Branch off https://github.com/uport-project/3box-js/pull/60

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -4,6 +4,7 @@ const localstorage = require('store')
 const IPFS = require('ipfs')
 const OrbitDB = require('orbit-db')
 const Room = require('ipfs-pubsub-room')
+const sha256 = require('js-sha256').sha256
 
 const PublicStore = require('./publicStore')
 const PrivateStore = require('./privateStore')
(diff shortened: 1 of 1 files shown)


zachferland:
Resolves https://github.com/uport-project/3box-js/issues/22

src/3box.js
```diff
@@ -173,7 +174,10 @@ class ThreeBox {
     if (serializedMuDID) {
       muportDID = new MuPort(serializedMuDID)
     } else {
-      const entropy = (await utils.openBoxConsent(address, web3provider)).slice(2, 34)
+      const sig = await utils.openBoxConsent(address, web3provider)
+      const entropy = sha256(sig.slice(2))
```
oed: The sha356Multihash is not used here because the entropy string becomes to long right?

zachferland: hmm actually didn't check if bip39.entropyToMnemonic require 32, it would for 12 word I guess. But mainly don't use it since it doesn't contribute any useful material as input to entropyToMnemonic, and doesn't seem to provide any useful intermediate step, unless if could in the future

oed: Yeah, that makes sense :)

src/3box.js
```diff
@@ -173,7 +174,10 @@ class ThreeBox {
...
+  // console.log(entropy)
+      // const entropy = (await utils.openBoxConsent(address, web3provider)).slice(2, 34)
```
zachferland: forgot to remove,  will do that

PR #64, opened 2018-10-02
Feat/address server sync

oed:
With this PR 3box should work correctly if used within the same dapp.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -27,11 +27,11 @@
       ThreeBox.openBox(web3.eth.accounts[0],  web3.currentProvider).then(box => {
         console.log(box)
 
-        updateProfileData(box)
         controlls.style.display = 'inline'
+        updateProfileData(box)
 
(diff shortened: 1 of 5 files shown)


oed:
Hm, why is codecov complaining?

zachferland:
@oed mostly name changes in this diff? Is it useful if we also run the example? 

Yeah don't get the code cov thing, says the diff alone is below target, even thought the entire branch still hits the exact same target, would probably ignore for now

oed:
@zachferland Yeah, so that we use the same names as the address-server does. Yeah try running the example, add an entry, reload the page and see if it shows up the second time.
Note that the example needs to be hosted on a server for MM to inject web3

zachferland:
@oed yup makes sense, tried the example, got a few 404s but all worked correctly (and couldn't get the same 404s again). 

For the example I can add a simple static local server as part of other example story, so that people can quickly run the example

PR #65, opened 2018-10-02
feat: add internal metadata to sets, timestamp

zachferland:
Adds timestamps to set data. Keeps for internal use for now. We may want to add other additional internal metadata in the future, this starts using it as { value: 'val set by app', timestamp: ..., metadata1: ..., metadata2: ....}. While a get only returns value.  Maybe want to externalize in future.

### DIFF
--- a/src/keyValueStore.js
+++ b/src/keyValueStore.js
@@ -15,7 +15,8 @@ class KeyValueStore {
    */
   async get (key) {
     if (!this._db) throw new Error('_sync must be called before interacting with the store')
-    return this._db.get(key)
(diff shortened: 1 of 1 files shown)


christianlundkvist:
@oed Having the timestamps there do reveal some information, but if someone really wanted to they could also just monitor the values in your private store and log when they change, so I don't think it's a major difference.

zachferland:
timestamps would also be encrypted with vals, are there other privacy preserving constraints defined for the private store right now? other than simply encrypting all vals

oed:
@christianlundkvist that's a good point. This PR should be good to go then!

PR #67, opened 2018-10-03
Feat/prep beta 5

oed:
This fixes a small bug in the .all function in the public store and bumps the beta version.

Already published to npm btw, b/c kenzo wanted this in the morning.


zachferland:
just beat me to it aha, saw that last night was just about to fix this morning, looks good

PR #68, opened 2018-10-03
feat: make available time order log/history of store

zachferland:
One problem related to this PR and the timestamp one is that remove or DEL events don't have timestamps, right not DEL events are still returned but with value null and timestamp null. Probably worth resolving after this.

### DIFF
--- a/src/__tests__/keyValueStore.test.js
+++ b/src/__tests__/keyValueStore.test.js
@@ -61,6 +61,52 @@ describe('KeyValueStore', () => {
     await ipfs2.stop()
   })
 
+  describe('log', () => {
+
+    beforeEach(async () => {
+      keyValueStore = new KeyValueStore(orbitdb, STORE_NAME)
+      storeAddr = await keyValueStore._sync()
+      await keyValueStore.set('key1', 'value1')
+      await keyValueStore.set('key2', 'lalalla')
+      await keyValueStore.set('key3', '12345')
+    })
+
+    it('should return array of ALL entries ({op: .., key: .., value: .., timestamp: ..}) of log underlying store ', async () => {
+      const log = keyValueStore.log
+      expect(log.length).toEqual(3)
+      const entry = log[0]
+      expect(Object.keys(entry).sort()).toEqual(['key', 'op', 'timestamp', 'value'])
+    })
+
+    it('should be time ordered', async () => {
+      const log = keyValueStore.log
+      expect(log[0].key).toEqual('key1')
+      expect(log[1].key).toEqual('key2')
+      expect(log[2].key).toEqual('key3')
+    })
+
+    it('should including ALL entries, including OPS on same keys', async () => {
+      // write over existing key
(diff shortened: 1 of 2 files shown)


src/__tests__/keyValueStore.test.js
```diff
@@ -61,6 +61,42 @@ describe('KeyValueStore', () => {
...
+    it('should including ALL entries, including OPS on same keys', async () => {
+      // write over existing key
+      await keyValueStore.set('key3', '6789')
+      const log = keyValueStore.log
+      expect(log[2].key).toEqual('key3')
+      expect(log[3].key).toEqual('key3')
+      expect(log[2].value).toEqual('12345')
+      expect(log[3].value).toEqual('6789')
+    })
+  })
```
oed: Maybe we should add a test for `DEL` events as well?

src/keyValueStore.js
```diff
@@ -80,6 +80,27 @@ class KeyValueStore {
...
+   *  // { op: 'PUT', key: 'Name', value: 'Botbot', timestamp: '1538575416068' }
+   *
+   * @return    {Array<Object>}     Array of ordered log entry objects
+   */
+  get log () {
+    return this._db._oplog.values.map(obj => {
+      return { op: obj.payload.op,
+        key: obj.payload.key,
+        value: obj.payload.value ? obj.payload.value.value : null,
+        timestamp: obj.payload.value ? obj.payload.value.timestamp : null }
```
oed: Does it make more sense to use the timestamp of the previous update rather than `null`?

zachferland:
Yeah we can do that, my preference would be to still squash/amend commits often right before merging, since while I agree additional commits are useful here in a pr/comments, I find the cleaned up commits easier to follow in the future. But whatever gets us in the same workflow, sounds good

oed:
@zachferland I haven't considered that usecase too much though, going back at a later point. Usually you can just separate stuff by merge right? Anyway, making a habit of doing "squash and merge" sounds good!

zachferland:
yeah if we use merge commits, I would often use the rebase and merge instead and try to minimize merge commits, to keep it all linear and readable, but yeah in that case the commits have to be useful/descriptive as well. I don't like the GitHub squash and merge option, but I like to cleanup up commits before merging (and squashing some). Lets keep your work flow for now, no reason to spend time on it right now.

oed:
What's the difference between squashing yourself and githubs squasing? (haven't used githubs). I definitely prefer merge commits, as it clearly shows what has happened.
Still wan't @christianlundkvist to lay an eye on this before merging :)

zachferland:
Realized another todo for this that would make sense, should extend in private store and return all vals unencrypted rather than encrypted at the moment

oed:
@zachferland yeah, good point should we add to this PR?

zachferland:
@oed yes, unless blocker for kenzo, then should merge and release so that he can try it. Then I can add this in another pr

PR #75, opened 2018-10-04
Replication working

c-castillo:
- Library uses `ipfs-pubsub-room` to broadcast the root store address to other nodes (we are interested the remote server receives the messages and open the orbitDB database. Local ipfs nodes don't poll for messages)
- Server endpoint added

(For the server code, take a look at the 3box-pinning-server repo)


c-castillo:
@oed can you help me figure out what's wrong with the tests? maybe some things will need to change

oed:
Thanks @c-castillo! The branch was a bit weird for some reason, so I squashed the commits. Also added tests for the pubsub stuff and did some cleanup.

zachferland:
Nice! approach seems to makes sense to me. I tried running the example from two different servers, and everything seems to work nicely. Noticed a bit of code comments that look like they will be resolved later. Is this the first approach? will there be other approaches later? or what are the improvements and concerns for later?

oed:
@zachferland I think we need to add some things to this approach where the pinning server tells the syncing client how many entries it has. It should also tell the client when it has finished replicating the db. This way we can get more security in that the replication is actually taking place as it should (and recover if it hasn't)

PR #79, opened 2018-10-05
Added consent callback function to openBox

oed:


src/3box.js
```diff
@@ -170,15 +170,18 @@ class ThreeBox {
...
    * @param     {String}        opts.orbitPath          A custom path for orbitdb storage
+   * @param     {Function}      opts.consentCallback    A function that will be called when the user has consented to opening the box
    * @return    {ThreeBox}                              the threeBox instance for the given address
    */
   static async openBox (address, web3provider, opts = {}) {
     let muportDID
     let serializedMuDID = localstorage.get('serializedMuDID_' + address)
     if (serializedMuDID) {
       muportDID = new MuPort(serializedMuDID)
+      if (opts.consentCallback) opts.consentCallback('user already logged in')
```
zachferland: do we differentiate here for UI reasons? maybe we pass a more standard val, while very dev readable not super machine readable. Maybe bool, or if eventually more values, some simpler const strings

oed: Yeah, this was a bit rushed. I guess a bool would make more sense for now

oed: I'm a bit unsure of how to document it though.

oed: I'm changing to true/false. Let's document this later.

Issue #82, opened 2018-10-08
Remove connection to bootstrap nodes


oed:
We are not currently using the bootstrap nodes, so connecting to them is just a waste of time. Perhaps we should put our ipfs node as one of the bootstrap nodes?

We should test to see what is quicker. Also test with and without "swarm" nodes. (see #91)

PR #91, opened 2018-10-08
Updating pinning server, swarm and bootstrap nodes

c-castillo:


src/3box.js
```diff
@@ -11,8 +11,27 @@ const OrbitdbKeyAdapter = require('./orbitdbKeyAdapter')
...
+    Bootstrap: [
+      '/dns4/ams-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd',
+      '/dns4/lon-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3',
+      '/dns4/sfo-3.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLPppuBtQSGwKDZT2M73ULpjvfd3aZ6ha4oFGL1KrGM',
+      '/dns4/sgp-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLSafTMBsPKadTEgaXctDQVcqN88CNLHXMkTNwMKPnu',
+      '/dns4/nyc-1.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmSoLueR4xBeUbY9WZ9xGUUxunbKWcrNFTDAadQJmocnWm',
+      '/dns4/wss0.bootstrap.libp2p.io/tcp/443/wss/ipfs/QmZMxNdpMkewiVZLMRxaNxUeZpDUb34pWjZ1kZvsd16Zic',
+      '/dns4/wss1.bootstrap.libp2p.io/tcp/443/wss/ipfs/Qmbut9Ywz9YEDrz8ySBSgWyJk41Uvm2QJPhwDJzJyGFsD6',
+      '/dns4/bootstrap.osliki.net/tcp/443/wss/ipfs/QmfJB77qXfiEdJkaSxpZgiMh9kAPiDBj3ga7TxF72QdWtf',
+      '/p2p-circuit/ipfs/QmfJB77qXfiEdJkaSxpZgiMh9kAPiDBj3ga7TxF72QdWtf'
```
oed: Do we need the bootstrap nodes? Why did you add them here?

src/3box.js
```diff
@@ -11,8 +11,27 @@ const OrbitdbKeyAdapter = require('./orbitdbKeyAdapter')
...
 
 const ADDRESS_SERVER_URL = 'https://beta.3box.io/address-server'
-const PINNING_SERVER = '/ip4/18.236.233.237/tcp/4003/ws/ipfs/QmbgpyTLCYBy84E1HGwei6niQHiLQmRpfN6SQtfJiyNMUd'
+const PINNING_SERVER = '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmbgpyTLCYBy84E1HGwei6niQHiLQmRpfN6SQtfJiyNMU'
 const IPFS_OPTIONS = {
+  config: {
+    Addresses: {
+      Swarm: [
+        '/dns4/wrtc-star.discovery.libp2p.io/tcp/443/wss/p2p-webrtc-star',
+        '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
```
oed: What are these? Looks like some webrtc enablers? Do you know what they do?

src/3box.js
```diff
@@ -11,8 +11,27 @@ const OrbitdbKeyAdapter = require('./orbitdbKeyAdapter')
 const utils = require('./utils')
 
 const ADDRESS_SERVER_URL = 'https://beta.3box.io/address-server'
-const PINNING_SERVER = '/ip4/18.236.233.237/tcp/4003/ws/ipfs/QmbgpyTLCYBy84E1HGwei6niQHiLQmRpfN6SQtfJiyNMUd'
+const PINNING_SERVER = '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmbgpyTLCYBy84E1HGwei6niQHiLQmRpfN6SQtfJiyNMU'
```
oed: How come we are using `/dnsaddr` rather than `/dns4` as the entries below?

src/3box.js
c-castillo: I preferred to whitelist them because one of them was throwing connection error (nyc-2 I believe)

oed: Oh, right. By default there are bootstrap nodes, but if we set this to an empty array we should not connect to any I think. see #82

src/3box.js
c-castillo: Only the first one seems to be a webrtc enabler. I think it's safe to remove them after testing they work. But I don't find harmful them to be there.

oed: let's keep for now, we can test in #82

src/3box.js
oed: Ok, strange that it works for the other bootstrap nodes 🤔

PR #92, opened 2018-10-09
Add keys used by attestari.me

joaoaguiam:
Feel free to place them on a different location, wasn't sure the best place to add those keys.


michaelsena:
Merged @joaoaguiam :+1:

joaoaguiam:
@oed I would propose to have a section per dapp, or a page per dapp and a link on this one. This way we can detail better each data structure without polluting this summary page. What do you think?

oed:
@joaoaguiam yes, that make sense. Let's add a section called **Entry data structure** and then a subsection for each entry type. We can then also link this entry from the entry name 🙂

PR #99, opened 2018-10-09
Fix/babel config es5

zachferland:
without any babel configs, nothing was transformed, input === output. This adds config for es5 and removes mockst/tests from lib output

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,5 +8,9 @@
   "plugins": [
     "syntax-object-rest-spread",
     "transform-object-rest-spread"
-  ]
+  ],
+  "ignore": [
+     "src/__mocks__",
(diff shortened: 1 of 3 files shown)


zachferland:
Babel recently made some breaking changes, causing a bunch of issues, so still trying to resolve the test failures, basically the build process works, but jest keeps trying to using another version of babel.

zachferland:
babel 7 breaking changes, babel 6 still in jest, plus other options for jest to ignore babelrc file not working and seems like if test are babelified they fail anyways, closing PR, quick solution with out babelrc file here https://github.com/uport-project/3box-js/pull/105

May still need to resolve, to get it work well with other build process like web dist, etc

PR #105, opened 2018-10-10
fix: add opt, so babel output transformed

zachferland:


zachferland:
Resolves https://github.com/uport-project/3box-js/pull/99 for now

PR #106, opened 2018-10-10
build: to webpack, add babel, min output

zachferland:
browserify to webpack (will allow more control over build issues)
es5 web build output

Minified output, results in error we likely will not be able to resolve. And since webpack makes the dist 3 mb large than browserify, it is not worth merging at the moment

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "3box",
-  "version": "1.0.0-beta-6",
+  "version": "1.0.0-beta-7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


zachferland:
All set now, no idea, but explicitly defining the same plugin webpack optimizer uses by default works

zachferland:
Still concerning that it doesn't work by default, there could be issues in others builds 
Also the build is 11.5mb, while the minified is 3.98 mb, both sizes are concerning

zachferland:
We should also remove the dist folder from github, just publish it with npm releases

oed:
That's a pretty good improvement though! I think we need to look more closely at our dependencies for further improvements. Yep, I think we have a story for a prepublish hook in zenhub for next sprint!

oed:
minified ipfs seems to be taking up 2.3 mb btw: https://bundlephobia.com/result?p=ipfs@0.32.0

oed:
One idea I have is to put the ipfs into an iframe on the 3box.io domain. then we can route the ipfs api though that ipfs object. We can use progressive webapp tech to make it so the user only have to load the ipfs library once. 
This summary is not so good, but I think all components are there. Let's sync on the details, but I guess is something for after the 1.0.0 release.

zachferland:
@oed yesss, I was thinking something along the exact same lines, not sure how much we will gain in the iframe since is still has to load everything there, but with separating ipfs libs will have some gains (like loading less often, more likely to be cached in a CDN, etc)

zachferland:
and yeah, lets move any additional improvement to later or other prs, this is likely the best we will get with current tooling optimizations

oed:
I was under the impression that you could store stuff locally using some tech for progressive webapps. If that is the case we could make sure that the iframe loads the ipfs lib locally. Or maybe what's in the iframe could be a service worker with an ipfs node. Don't know the details of how this would work though.

zachferland:
ah in the sense, yeah might work, Im not familiar enough yet, but worth looking into more

oed:
@zachferland are you adding the prepublish hook to this PR or another one?

PR #111, opened 2018-10-11
Remove webrtc swarm addr

oed:

PR #112, opened 2018-10-11
Feat/fix publicstore index bug

oed:
Found this bug as I was playing around with the box object and `all()` function in publicStore. This function had side effects on the internal representation of the db in orbitdb-kvstore.

###### NOTE: Merge this after #111


zachferland:
hmm would not have expected that, but yeah best to return new obj, if you want to make it more functional and once line, since already using map, could do 
```js
Object.keys(entries).reduce((acc, key) => { acc[key] = entries[key].value; return acc }, {})
```

PR #114, opened 2018-10-11
docs: add shields and menu

zachferland:


zachferland:
@michaelsena looks like you have enable something/widget for the discord shield to show some stats

michaelsena:
I enabled the widget @zachferland, is there some URL or code needed for it to pull member numbers?

zachferland:
@michaelsena prefer member number vs number online? can see if that is an option

zachferland:
nvm, this is the only option for now

PR #116, opened 2018-10-11
ref: sync DBs as soon as known

zachferland:
Didn't compare times yet, but looks to be very small, nonetheless, helps generalize some our own handling all these events. Also don't quite understand when the the replicated event is pushed, comment left in code (and commented code).

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -117,30 +117,29 @@ class ThreeBox {
       // console.log(this._rootStore.iterator({ limit: -1 }).collect().length)
       // console.log(await this._ipfs.pubsub.peers('/orbitdb/QmRxUAGk62v7NjUkzvcqwYkBqF3zHb8tfhfW6T3MateGje/b932fe7ab.root'))
       console.time('replicate rootStore')
-      if (!this._rootStore.iterator({ limit: -1 }).collect().length) {
-        await new Promise((resolve, reject) => {
-          this._rootStore.events.on(
-            'replicate.progress',
-            (_x, _y, _z, num, max) => {
-              if (num === max) {
-                console.timeEnd('replicate rootStore')
-                this._rootStore.events.on('replicated', resolve)
-              }
-            }
-          )
-        })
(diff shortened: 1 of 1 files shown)


src/3box.js
```diff
@@ -117,30 +117,40 @@ class ThreeBox {
       // console.log(this._rootStore.iterator({ limit: -1 }).collect().length)
       // console.log(await this._ipfs.pubsub.peers('/orbitdb/QmRxUAGk62v7NjUkzvcqwYkBqF3zHb8tfhfW6T3MateGje/b932fe7ab.root'))
       console.time('replicate rootStore')
-      if (!this._rootStore.iterator({ limit: -1 }).collect().length) {
+      const entries = this._rootStore.iterator({ limit: -1 }).collect()
+      const rootStoreEmpty = !entries.length
+      // const rs = this._rootStore
+      // function replicatedRootStore () {
+      //   return new Promise((resolve, reject) => {
+      //     rs.events.on('replicated', () => {
```
oed: The `'replicated'` event is emitted when an entry has been received **and** fully added to the local odb.

oed: I asked haad about this. I though this would only be emitted when the entire store had been replicated, but turns out it's emitted for every event.

src/3box.js
```diff
@@ -117,30 +117,40 @@ class ThreeBox {
...
+        // storePromises.push(replicatedRootStore())  // would have thought this did that same as wrapping below in promise??
         await new Promise((resolve, reject) => {
-          this._rootStore.events.on(
-            'replicate.progress',
-            (_x, _y, _z, num, max) => {
-              if (num === max) {
-                console.timeEnd('replicate rootStore')
-                this._rootStore.events.on('replicated', resolve)
-              }
+          this._rootStore.events.on('replicate.progress', (address, hash, entry, progress, have) => {
```
oed: The `'replicate.progress'` event is emitted as soon as an event has been received from the network, but not added to the local odb yet.

PR #135, opened 2018-10-18
feat: use new pubsub response, dont wait for root address

zachferland:
Need to try a different approach here, create root address, then can't determine if first time from root address, luckily new pubsub responses can be used, but then this is just slower most of the time. So need think about further and reorg. Will pick up tmrrw again

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -13,7 +13,6 @@ const utils = require('./utils')
 
 const ADDRESS_SERVER_URL = 'https://beta.3box.io/address-server'
 const PINNING_NODE = '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmZvxEpiVNjmNbEKyQGvFzAY1BwmGuuvdUTmcTstQPhyVC'
-// const PINNING_NODE = '/ip4/127.0.0.1/tcp/4003/ws/ipfs/QmNxwyrobAM9rWAvzCC4QKyvRvadUwsQtGS9cqz6hVRc3a'
 const PINNING_ROOM = '3box-pinning'
 const IPFS_OPTIONS = {
   EXPERIMENTAL: {
@@ -42,7 +41,7 @@ class ThreeBox {
     this.private = null
   }
 
-  async _load (opts = {}) {
+  async _load (syncComplete, opts = {}) {
     const did = this._muportDID.getDid()
     const didFingerprint = utils.sha256Multihash(did)
     const rootStoreName = didFingerprint + '.root'
@@ -79,22 +78,30 @@ class ThreeBox {
       }
     }
 
-    this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
-
     this.public = new PublicStore(this._orbitdb, didFingerprint + '.public', this._linkProfile.bind(this))
     this.private = new PrivateStore(this._muportDID, this._orbitdb, didFingerprint + '.private')
 
     console.time('init stores')
-    const [pubOdbAddress, privOdbAddress] = await Promise.all([this.public._init(), this.private._init()])
+    const [pubStoreAddress, privStoreAddress] = await Promise.all([this.public._init(), this.private._init()])
     console.timeEnd('init stores')
 
-    this._createRootStore(rootStoreAddress, privOdbAddress, pubOdbAddress, pinningNode)
-  }
+    let syncPromises = []
 
-  async _sync () {
-    console.time('sync stores')
-    await Promise.all([this.public._sync(), this.private._sync()])
-    console.timeEnd('sync stores')
+    const onMessageRes = (topic, data) => {
+      if (data.type === 'HAS_ENTRIES') {
+        if(data.odbAddress === privStoreAddress) {
+          syncPromises.push(this.private._sync(data.numEntries))
+        }
+        if(data.odbAddress === pubStoreAddress) {
+          syncPromises.push(this.public._sync(data.numEntries))
+        }
+        if (syncPromises.length === 2) Promise.all(syncPromises).then(syncComplete)
+      }
+    }
+
+    this._pubsub.subscribe(PINNING_ROOM, onMessageRes, onNewPeer)
+
+    this._createRootStore(rootStoreAddress, privStoreAddress, pubStoreAddress, pinningNode)
   }
 
   async _createRootStore(rootStoreAddress, privOdbAddress, pubOdbAddress) {
@@ -190,7 +197,7 @@ class ThreeBox {
    * @param     {Function}      opts.consentCallback    A function that will be called when the user has consented to opening the box
    * @return    {ThreeBox}                              the threeBox instance for the given address
    */
-  static async openBox (address, web3provider, opts = {}) {
+  static async openBox (address, web3provider, syncComplete, opts = {}) {
     console.time('-- openBox --')
     let muportDID
     let serializedMuDID = localstorage.get('serializedMuDID_' + address)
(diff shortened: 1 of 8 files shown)


src/3box.js
```diff
@@ -78,53 +93,51 @@ class ThreeBox {
...
+        this._rootStore.events.on('replicate.progress', (address, hash, entry, progress, have) => {
+          storePromises.push(dbSyncPromise(entry))
+          if (progress === have) {
+            resolve()
+            console.timeEnd('replicate rootStore')
+          }
         })
-      } else {
+      })
+    } else if (!rootStoreEmpty) {
```
oed: If this happens we shouldn't wait for the `replicatedRootStore` promise.

src/3box.js
```diff
@@ -134,14 +147,16 @@ class ThreeBox {
       console.time('add PrivateStore to rootStore')
       await this._rootStore.add({ odbAddress: await this.private._sync() })
       console.timeEnd('add PrivateStore to rootStore')
-      rootStoreAddress = this._rootStore.address.toString()
+      // rootStoreAddress = this._rootStore.address.toString()
       console.time('publish rootStoreAddress to address-server')
       this._publishRootStore(rootStoreAddress)
       console.timeEnd('publish rootStoreAddress to address-server')
       console.time('broadcast rootStoreAddress over pubsub')
       this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
```
oed: This line isn't really needed because we do it on line 87.

src/3box.js
```diff
@@ -66,10 +71,20 @@ class ThreeBox {
...
-        this._pubsub.publish(PINNING_ROOM, { odbAddress: rootStoreAddress })
+        // TODO add type   type: 'PIN_DB',
+        this._pubsub.publish(PINNING_ROOM, {type: 'PIN_DB', odbAddress: rootStoreAddress })
+        // TODO subscribe to wait for response, to get length
       }
     }
-    if (rootStoreAddress) this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
+
+    const entriesNum = (resolve) => (topic, data) => {
+      return data.type === 'HAS_ENTRIES' ? resolve(data.numEntries) : resolve(null)
```
oed: We also need to check that `data.odbAddress === rootStoreAddress`

src/3box.js
```diff
@@ -78,53 +93,51 @@ class ThreeBox {
     globalIPFS = this._ipfs
     globalOrbitDB = this._orbitdb
 
+    this._rootStore = await this._orbitdb.feed(rootStoreName)
+    const rootStoreAddress = this._rootStore.address.toString()
```
oed: I guess these two lines needs to be moved to above the pubsub stuff.

zachferland:
thanks @oed for the quick review already, I hadn't even reviewed it yet :) Going to do some cleanup and those fixes

zachferland:
Removes lot of code, that we will likely have to add back later in some form. Haven't tested through everything with the pinning server issues.

src/3box.js
```diff
@@ -66,82 +76,55 @@ class ThreeBox {
...
-    if (rootStoreAddress) this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
 
-    const keystore = new OrbitdbKeyAdapter(this._muportDID)
-    console.time('new OrbitDB')
-    this._orbitdb = new OrbitDB(this._ipfs, opts.orbitPath, { keystore })
-    console.timeEnd('new OrbitDB')
-    globalIPFS = this._ipfs
-    globalOrbitDB = this._orbitdb
+    const parsePubSubMessage = (resolve) => (topic, data) => {
+      if (data.type === 'HAS_ENTRIES' & data.odbAddress === rootStoreAddress) {
```
oed: I'd prefer to use a logical and (&&) instead of a bitwise and.

src/keyValueStore.js
```diff
@@ -71,11 +71,16 @@ class KeyValueStore {
...
     } else {
-      this._db = await this._orbitdb.keyvalue(this._name)
+      this._init()
     }
     return this._db.address.toString()
   }
 
+  async _init() {
+    this._db = await this._orbitdb.keyvalue(this._name)
+    return this._db.address.toString()
```
oed: Let's do this._db.load() here ans await the `'ready'` event.

zachferland:
great, yeah didn't do it on the assumption of having onUpdate, sounds good to just start here

src/3box.js
zachferland: yup definitely

src/keyValueStore.js
zachferland: 👍

oed:
@zachferland only on remote updates was my thinking. Lets sync on this after the standup!

zachferland:
also going to offer some indication or when remote is fully synced? like an onSyncComplete event or something, for anyone who doesn't mind waiting to read until all data is available

zachferland:
:+1 yup lets sync after standup

src/3box.js
```diff
@@ -57,91 +56,61 @@ class ThreeBox {
...
-      rootStoreAddress = this._rootStore.address.toString()
-      console.time('publish rootStoreAddress to address-server')
-      this._publishRootStore(rootStoreAddress)
-      console.timeEnd('publish rootStoreAddress to address-server')
-      console.time('broadcast rootStoreAddress over pubsub')
-      this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
-      console.timeEnd('broadcast rootStoreAddress over pubsub')
     }
+    console.time('open rootStore')
+    this._rootStore = await this._orbitdb.open(rootStoreAddress)
```
oed: _rootStore is already created at line 66

src/3box.js
```diff
@@ -57,91 +56,61 @@ class ThreeBox {
...
+    await this._rootStore.add({ odbAddress: pubOdbAddress })
+    console.timeEnd('add PublicStore to rootStore')
+    console.time('add PrivateStore to rootStore')
+    await this._rootStore.add({ odbAddress: privOdbAddress })
+    console.timeEnd('add PrivateStore to rootStore')
+    console.time('publish rootStoreAddress to address-server')
+    this._publishRootStore(rootStoreAddress)
+    console.timeEnd('publish rootStoreAddress to address-server')
+    console.time('broadcast rootStoreAddress over pubsub')
+    this._pubsub.subscribe(PINNING_ROOM, () => {}, onNewPeer)
```
oed: It might be a good idea to do this on line 99 or so. It isn't blocking anything else, and if it happens earlier the syncs will be faster!

src/keyValueStore.js
```diff
@@ -71,11 +70,17 @@ class KeyValueStore {
         })
       })
     } else {
-      this._db = await this._orbitdb.keyvalue(this._name)
+      this._init() //TODO probably don't need this
```
oed: Yep, this can be removed

src/keyValueStore.js
```diff
@@ -51,7 +51,6 @@ class KeyValueStore {
       const readyPromise = new Promise((resolve, reject) => {
         this._db.events.on('ready', resolve)
       })
-      this._db.load()
       await readyPromise
```
oed: Since we are awaiting `.load()` we don't need the readyPromise at all.

src/3box.js
```diff
@@ -57,91 +56,61 @@ class ThreeBox {
...
+    console.timeEnd('init stores')
+
+    this._createRootStore(rootStoreAddress, privOdbAddress, pubOdbAddress, pinningNode)
+
+    return {rootStoreAddress, privOdbAddress, pubOdbAddress}
+  }
+
+  async _sync (privOdbAddress, pubOdbAddress) {
+    console.time('sync stores')
+    await Promise.all([this.public._sync(pubOdbAddress), this.private._sync(privOdbAddress)])
```
oed: We don't really need the odbAddresses here. the `_sync()` method should assume that `_init()` has been called and can therefore use `this._db` directly.

src/3box.js
```diff
@@ -223,12 +226,21 @@ class ThreeBox {
...
     console.timeEnd('-- openBox --')
     return threeBox
   }
 
+  /**
+   * Sets the callback function that will be called once when the db is fully synced.
+   *
+   * @param     {Function}      syncDone        the function that will be called
+   */
+  onSyncDone (syncDone) {
```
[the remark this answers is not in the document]
oed: I don't think that's very likely though. Do you feel like that could be a real concern?

PR #146, opened 2018-10-23
feat: use 3box pinning server for muport

zachferland:


src/3box.js
```diff
@@ -14,6 +14,7 @@ const utils = require('./utils')
 const ADDRESS_SERVER_URL = 'https://beta.3box.io/address-server'
 const PINNING_NODE = '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmZvxEpiVNjmNbEKyQGvFzAY1BwmGuuvdUTmcTstQPhyVC'
 const PINNING_ROOM = '3box-pinning'
+const IPFS_MUPORT = { protocol: 'https', host: 'ipfs.3box.io' }
```
oed:
```suggestion
const IPFS_MUPORT = { protocol: 'https', host: 'ipfs.3box.io', port: 5002 }
```

oed:
js-ipfs doesn't support the `/cat/` endpoint used by ipfs-mini (which is in turn used by muport). This unfortunately means that this won't work at all. 
In the future we should instead look into using an ipfs instance in muport-core and muport-did-resolver directly. This would allow us to do something similar to this, but more decentralized :D

PR #147, opened 2018-10-23
build: up to babel 7

zachferland:


oed:
@zachferland assigned you and estimated as a 2.

PR #157, opened 2018-10-25
Deterministically generate salt

oed:


src/privateStore.js
```diff
@@ -46,17 +47,9 @@ class PrivateStore extends KeyValueStore {
...
-    if (encryptedSalt) {
-      this._salt = this._decryptEntry(encryptedSalt)
-    } else {
-      this._salt = Buffer.from(nacl.randomBytes(16)).toString('hex')
-      encryptedSalt = this._encryptEntry(this._salt)
-      await super.set(SALT_KEY, encryptedSalt)
-    }
-    return address
+  _generateSalt () {
+    const saltData = this.muportDID.keyring.signingKey.deriveChild(0)._hdkey._privateKey.toString('hex')
```
zachferland: how we derive multiple salts for multiple private stores or name spaces? or not really a concern until we get there? And the intention in our use is to keep the salt private right? (vs public, since serving a different purpose here, and keys are likely known vals)

oed: I imagine that we can do something like `store-name % n`. Yeah, the idea is to keep it private :)

PR #164, opened 2018-10-29
Added example for getProfile

oed:


zachferland:
If merging into master, have to also merge into develop after, may be easier to just merge into develop

oed:
Oh, yes this should be merged into develop. My bad!

PR #165, opened 2018-11-03
Publish the dist directory to npm, not lib

tyleryasaka:
I tried requiring 3box in a project, and got a bunch of babel errors. I realized that this was because the `3box.js` file in the `lib` directory had been published to npm. The compiled file of course, is in `dist/3box.js`.

Maybe there's a good reason for this that I'm not aware of - but as it currently is I can't successfully require 3box in my project.

Example error message (I got over a dozen like this):
```
3box.js:3 Uncaught Error: Cannot find module '@babel/runtime/helpers/interopRequireDefault'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "3box",
   "version": "1.0.3",
   "description": "Interact with user data",
-  "main": "lib/3box.js",
+  "main": "dist/3box.js",
   "directories": {
     "lib": "lib"
   },


michaelsena:
Thanks for letting us know @tyleryasaka. @oed or @zachferland would you guys mind taking a look?

oed:
@tyleryasaka Does it work when you switch to the dist file? The dist file is the browserified file intended to be used in a html `<script>` tag directly. The file in `lib` is a build to es5 while the source code in `src` is written with es6. The reason `lib` is main is that your projects build setup should be able to optimize the resulting bundle better. (correct me if I'm wrong @zachferland) 

How do you build you project btw?

zachferland:
yup what @oed said is true

The purpose of the lib folder (being es5) is to be widely compatible across browser versions and can be used as is in your own build tool. No transpiling is necessary, usually don't have to transpile dependencies in you own build process. It is assumed that if you are requiring/importing the dependency you must have some js build/bundles process in your app. Otherwise if you want to just use the lib (already bundled by it self) on some page in your app you would use unpkg.com or host the dist file. 

We wouldn't want to include the dist file as entry since that is already built/bundled wepback. Making it too opinionated and possibly inefficient in other build process. For example using the dist file in another webpack build wouldn't allow it to tree shake dependencies, so you could be bundling dependencies redundantly. This is especially a lot of overhead when dealing with large dependencies like ipfs with close to 2mb. 

hmm besides all that there seems to be some issue, you shouldn't have to include babel, or an additional dependency, and you shouldn't have to perform any webpack transformations on the package. Maybe there is some babel polyfill that is including a babel dependency for some reason. I'll have to look into it further, thanks @tyleryasaka for sharing the issue!

zachferland:
ah actually found the answer to my own question pretty quickly, '@babel/runtime' is a required production dependency, which makes sense (https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html), @oed  so yes, we will have to add that to our package dependencies 

yup and as @oed said already installing `@babel/runtime` is the easiest work around

zachferland:
https://github.com/uport-project/3box-js/pull/166

tyleryasaka:
@oed @zachferland thanks! Installing `@babel/runtime` fixed it.

That makes sense about the lib folder.

PR #170, opened 2018-11-08
Always lower-case ethereum addresses that are function inputs

### LINKED ISSUE #169: Different orbit-db instances in my Dapp and 3box-dapp
Note: I already have a fix in: https://github.com/uport-project/3box-js/pull/170

**Describe the bug**

From discord:

> I've been manipulating data through 3box-js for the last couple of days. Then I logged into 3box.io and added some info there.
> 
> Now for the weird part. When running my local 3box-js, box.public.get('some-key') returns a different value than Box.getProfile(...). Box.getProfile(...) returns the values I set in 3box.io, while box.public returns the values I've been manipulating locally the past couple of days.

**To Reproduce**
1. Set your profile name on 3box.io (Rinkeby is fine)
2. (In your Dapp that is using 3box-js) Pass in a checksummed or upper-cased address to the `openBox` method.
2. Try to get your profile name using `public.get('name')`
3. Your name should not be there. In fact, no data from 3box.io should be retrievable from `public.get`.

**Expected behavior**
Data should be synced between custom Dapp using 3box-js and 3box.io, for any given wallet.

**I figured out the problem**
The addresses are not checksummed in 3box-dapp, but in my dapp they were checksummed. A simple solution would be to lowercase all ethereum addresses that are inputs for functions in 3box-js.

Proposed fix: https://github.com/uport-project/3box-js/pull/170

(Tested, and this fixes the problem for me.)

tyleryasaka:
This way we're not making any assumptions about whether the addresses have been checksummed or not.

This is backwards-compatible with existing profiles created using 3box-dapp, as the addresses being used are already lower-cased.

Resolves #169

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -208,26 +208,27 @@ class Box {
    * @return    {Box}                                       the 3Box instance for the given address
    */
   static async openBox (address, ethereumProvider, opts = {}) {
+    const normalizedAddress = address.toLowerCase()
     // console.time('-- openBox --')
     let muportDID
-    let serializedMuDID = localstorage.get('serializedMuDID_' + address)
+    let serializedMuDID = localstorage.get('serializedMuDID_' + normalizedAddress)
     if (serializedMuDID) {
       // console.time('new Muport')
       muportDID = new MuPort(serializedMuDID)
       // console.timeEnd('new Muport')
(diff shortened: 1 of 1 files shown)


zachferland:
Thanks @tyleryasaka

PR #178, opened 2018-11-14
fix: connect to pin node directly, libp2p discovery servers not needed

zachferland:
Also connect to pinning node for getProfile. libp2p discovery nodes were throwing some errors the last few days, so lib could not find our ipfs node and sync the data.


c-castillo:
@zachferland :shipit:

PR #180, opened 2018-11-15
Fixed bug in getProfile when running node

oed:
Turns out XMLHttpRequest only returns `responseText` while browsers return both that and `response`.


oed:
codecov plz

PR #181, opened 2018-11-15
Handle _linkProfile errors correctly

oed:
The _linkProfile method was the only one that really needed updating. Now it only stores the link to localstorage if the call is successful.

### DIFF
--- a/src/__tests__/3box.test.js
+++ b/src/__tests__/3box.test.js
@@ -49,6 +49,7 @@ jest.mock('../utils', () => {
   const sha256 = require('js-sha256').sha256
   let addressMap = {}
   let linkmap = {}
+  let firstLink = true
   return {
     openBoxConsent: jest.fn(async () => '0x8726348762348723487238476238746827364872634876234876234'),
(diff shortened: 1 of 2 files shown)


src/__tests__/3box.test.js
```diff
@@ -61,10 +62,15 @@ jest.mock('../utils', () => {
           addressMap[did] = hash
           return { status: 'success', data: { hash } }
         case 'link': // make a link
-          did = payload.consent_msg.split(',')[1]
-          const address = payload.consent_signature.split(',')[1]
-          linkmap[address] = did
-          return { status: 'success', data: { did, address } }
+          if (firstLink) {
+            firstLink = false
+            return Promise.reject('{ status: "error", message: "an error" }')
```
c-castillo: Is this error message used internally only? Would it make more sense if the message is more descriptive?

oed: This is just the mock. The real error would come from the address-server, but for our testing purposes it doesn't really matter what it says.

PR #185, opened 2018-11-21
feat: add iframe store services connection, allows shared store acros…

zachferland:
### What 
This PR allows stores/data to be share across domains, resulting in less syncs and quicker starts. Initial attempts saw openBox time decrease by more than 50% when using same size data set

* 4-6 secs (when not opened on a domain yet) decreasing to 2.3-2.4  💨 🎉 
*  3 secs (when already cached on a domain) decreasing to 2.3-2.4 (idk why this changed yet 🤷‍♂️) 

### How 
Stores are shared across domains by running the storage layers in an iframe and using an RPC interface to connect the 3box library to the storage services running in the iframe. The iframe runs 'js-ipfs'  and [orbit-db-cache](https://github.com/orbitdb/orbit-db-cache) as a service. Then a proxy instance of both the cache and ipfs is used in 3box-js using [orbit-db-cache-postmsg-proxy](https://github.com/3box/orbit-db-cache-postmsg-proxy) and [ipfs-postmsg-proxy](https://github.com/ipfs-shipyard/ipfs-postmsg-proxy).

The iframe can be found here [3box-js-iframe](https://github.com/3box/3box-js-iframe). 

### Related

This PR is also connected to the following new repos and PRs:
* [orbit-db-cache-postmsg-proxy](https://github.com/3box/orbit-db-cache-postmsg-proxy)
* [3box-js-iframe](https://github.com/3box/3box-js-iframe)
* [PR for Cache option in OrbitDB](https://github.com/orbitdb/orbit-db/pull/507)

### Remaining TODOs:

- [x] Test in Node and make sure it also works there. Goal was to keep iframe optional, need better default when not using iframe. i.e provide ipfs, but remove from webpack build for browser build
- [x] Also iframe optional in browser?
- [ ] ~~Error thrown after sync, have not looked into where it is coming from yet~~
- [ ]~~Could Iframe load be faster? at least initial page loads faster now due to much smaller lib size, but then lags for iframe load.~~
- [x] ~~Maybe allow iframe path as option. Difficult because we want to load the iframe on page load since takes a little while. While passing an option would require for it to wait until it is called by app using lib. (Hard set to http://localhost:30001/ for testing right now)~~ Will be task in future, making iframe optional is sufficient for now.
- [ ] ~~Deploy iframe [3box-js-iframe](https://github.com/3box/3box-js-iframe). Could be at a space on our domain or on ipfs. Although we would likely want and a fixed reference. So that iframe can be updated without the all the libs needing updates.~~
- [x] Tests? Current test failure will be fixed, once first TODO is covered 
- [x] Change example. Either need how to on how to also run iframe locally, reference where ever we deploy it, or related 2nd TODO, run without iframe. 
- [ ] ~~Test once iframe deployed, to make sure there is no cross domain issues.~~

### Try it out

In this repo: 
```bash
npm run example:start
```

In [3box-js-iframe](https://github.com/3box/3box-js-iframe) repo at the same time:
```bash
npm run example:start
```

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -7,6 +7,12 @@ bopen.addEventListener('click', event => {
     preload: { enabled: false }
   }
 
+  /**
+   *  This uses a local ipfs-js instance and local orbitdb cache instead of the
+   *  shared iframe storage available at 3box.io. Easier for testing and differing
+   *  configs. But in production you will get the best performance by using the
+   *  default iframe configuration.
+   */
   const ipfs = new window.Ipfs(IPFS_OPTIONS)
   const opts = { ipfs, iframeStore: false }
 
@@ -14,6 +20,7 @@ bopen.addEventListener('click', event => {
     console.log('Sync Complete')
     updateProfileData(window.box)
   }
+
   window.ethereum.enable().then(addresses => {
     Box.openBox(addresses[0],  window.ethereum, opts).then(box => {
       box.onSyncDone(syncComplete)
--- a/example/index.js
+++ b/example/index.js
@@ -1,10 +1,21 @@
 bopen.addEventListener('click', event => {
+
+  const IPFS_OPTIONS = {
+    EXPERIMENTAL: {
+      pubsub: true
+    },
+    preload: { enabled: false }
+  }
+
+  const ipfs = new window.Ipfs(IPFS_OPTIONS)
+  const opts = { ipfs, iframeStore: false }
+
   const syncComplete = (res) => {
     console.log('Sync Complete')
     updateProfileData(window.box)
   }
   window.ethereum.enable().then(addresses => {
-    Box.openBox(addresses[0],  window.ethereum).then(box => {
+    Box.openBox(addresses[0],  window.ethereum, opts).then(box => {
       box.onSyncDone(syncComplete)
       window.box = box
       console.log(box)
@@ -44,7 +55,7 @@ bopen.addEventListener('click', event => {
 
 getProfile.addEventListener('click', () => {
   console.log(ethAddr.value)
-  Box.getProfile(ethAddr.value).then(profile => {
+  Box.getProfile(ethAddr.value, opts).then(profile => {
     console.log(profile)
     Object.entries(profile).map(kv => {
       getProfileData.innerHTML +=kv[0] + ': ' + kv[1] + '<br />'
--- a/example/index.js
+++ b/example/index.js
@@ -1,28 +1,12 @@
 bopen.addEventListener('click', event => {
 
-  const IPFS_OPTIONS = {
-    EXPERIMENTAL: {
-      pubsub: true
-    },
-    preload: { enabled: false }
-  }
-
-  /**
-   *  This uses a local ipfs-js instance and local orbitdb cache instead of the
-   *  shared iframe storage available at 3box.io. Easier for testing and differing
-   *  configs. But in production you will get the best performance by using the
-   *  default iframe configuration.
-   */
-  const ipfs = new window.Ipfs(IPFS_OPTIONS)
-  const opts = { ipfs, iframeStore: false }
-
   const syncComplete = (res) => {
     console.log('Sync Complete')
     updateProfileData(window.box)
   }
 
   window.ethereum.enable().then(addresses => {
-    Box.openBox(addresses[0],  window.ethereum, opts).then(box => {
+    Box.openBox(addresses[0],  window.ethereum, {}).then(box => {
       box.onSyncDone(syncComplete)
       window.box = box
       console.log(box)
@@ -62,7 +46,7 @@ bopen.addEventListener('click', event => {
 
 getProfile.addEventListener('click', () => {
   console.log(ethAddr.value)
-  Box.getProfile(ethAddr.value, opts).then(profile => {
+  Box.getProfile(ethAddr.value, {}).then(profile => {
     console.log(profile)
     Object.entries(profile).map(kv => {
       getProfileData.innerHTML +=kv[0] + ': ' + kv[1] + '<br />'
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ipfs-postmsg-proxy": "^3.1.1",
     "js-sha256": "^0.9.0",
     "muport-core": "^0.4.1",
-    "orbit-db": "^0.19.9",
+    "orbit-db": "git://github.com/orbitdb/orbit-db.git#dddb271",
     "orbit-db-cache-postmsg-proxy": "0.0.4",
     "store": "^2.0.12",
     "xmlhttprequest": "^1.8.0"
--- a/package.json
+++ b/package.json
@@ -43,13 +43,10 @@
     "@babel/runtime": "^7.1.2",
     "bip39": "^2.5.0",
     "ipfs": "^0.32.2",
-    "ipfs-api": "^24.0.0",
     "ipfs-postmsg-proxy": "^3.1.1",
-    "ipfs-pubsub-room": "^1.4.0",
     "js-sha256": "^0.9.0",
     "muport-core": "^0.4.1",
     "orbit-db": "^0.19.9",
-    "orbit-db-cache": "^0.2.4",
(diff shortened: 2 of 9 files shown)


zachferland:
### What 
This PR allows stores/data to be share across domains, resulting in less syncs and quicker starts. Initial attempts saw openBox time decrease by more than 50% when using same size data set

* 4-6 secs (when not opened on a domain yet) decreasing to 2.3-2.4  💨 🎉 
*  3 secs (when already cached on a domain) decreasing to 2.3-2.4 (idk why this changed yet 🤷‍♂️) 

### How 
Stores are shared across domains by running the storage layers in an iframe and using an RPC interface to connect the 3box library to the storage services running in the iframe. The iframe runs 'js-ipfs'  and [orbit-db-cache](https://github.com/orbitdb/orbit-db-cache) as a service. Then a proxy instance of both the cache and ipfs is used in 3box-js using [orbit-db-cache-postmsg-proxy](https://github.com/3box/orbit-db-cache-postmsg-proxy) and [ipfs-postmsg-proxy](https://github.com/ipfs-shipyard/ipfs-postmsg-proxy).

The iframe can be found here [3box-js-iframe](https://github.com/3box/3box-js-iframe). 

### Related

This PR is also connected to the following new repos and PRs:
* [orbit-db-cache-postmsg-proxy](https://github.com/3box/orbit-db-cache-postmsg-proxy)
* [3box-js-iframe](https://github.com/3box/3box-js-iframe)
* [PR for Cache option in OrbitDB](https://github.com/orbitdb/orbit-db/pull/507)

### Remaining TODOs:

- [ ] Test in Node and make sure it also works there. Goal was to keep iframe optional, need better default when not using iframe. i.e provide ipfs, but remove from webpack build for browser build
- [ ] Also iframe optional in browser?
- [ ] Error thrown after sync, have not looked into where it is coming from yet
- [ ] Could Iframe load be faster? at least initial page loads faster now due to much smaller lib size, but then lags for iframe load.
- [ ] Maybe allow iframe path as option. Difficult because we want to load the iframe on page load since takes a little while. While passing an option would require for it to wait until it is called by app using lib. (Hard set to http://localhost:30001/ for testing right now)
- [ ] Deploy iframe [3box-js-iframe](https://github.com/3box/3box-js-iframe). Could be at a space on our domain or on ipfs. Although we would likely want and a fixed reference. So that iframe can be updated without the all the libs needing updates. 
- [ ] Tests? Current test failure will be fixed, once first TODO is covered 
- [ ] Change example. Either need how to on how to also run iframe locally, reference where ever we deploy it, or related 2nd TODO, run without iframe. 
- [ ] Test once iframe deployed, to make sure there is no cross domain issues.

### Try it out

In this repo: 
```bash
npm run example:start
```

In [3box-js-iframe](https://github.com/3box/3box-js-iframe) repo at the same time:
```bash
npm run example:start
```

oed:
👍  Some questions for your todo list 🙂 

> Also iframe optional in browser?

I'm a bit torn on this. Maybe we can provide a way to build it in this way if desired?

> Error thrown after sync, have not looked into where it is coming from yet

In browser?

> Maybe allow iframe path as option. Difficult because we want to load the iframe on page load since takes a little while. While passing an option would require for it to wait until it is called by app using lib. (Hard set to http://localhost:30001/ for testing right now)

Maybe we can have a default that loads anyway, then if an option is passed change to that?

> Deploy iframe 3box-js-iframe. Could be at a space on our domain or on ipfs. Although we would likely want and a fixed reference. So that iframe can be updated without the all the libs needing updates.

Why do you think that being able to update it without updating all the libs? In my mind it makes sense to keep this as a part of the lib that gets updated with it.

>

zachferland:
@c-castillo the iframe (html file and js file) code that need to be deployed somewhere can be found here [3box-js-iframe](https://github.com/3box/3box-js-iframe)

zachferland:
@oed follow up
1) yeah can actually make it optional pretty easily, after revisiting it, basically going to remove ipfs dependency from libs, but if don't want to use iframe, dev just has to pass in correctly initialized/configure js-ipfs instance.
2) yeah in browser
3) yeah I was thinking something like that, but I am thinking now this is a later feature not part of this PR, as long as 1 (above is resolved) there is enough configurability 
4) Can't really decide on this, fixed ref is nice for planned updates, not so nice for unplanned updates (basically bugs). Any bugs in the iframe would force devs to update lib, and I am just a bit unsure of the stability of the libraries in the iframe yet, so not sure if there will be issues there. The plus is, our code there is pretty minimal, likely reducing any issue. If fixed ref, still would be versioned, just meant to say, not content addressed. Another option here is just release 3box-js-iframe as an npm package, then it will be available and versioned on CDNs like unpkg.com (cc @c-castillo )

oed:
1. That's a nice idea, like that!
4. I kind of feel that it would be simpler to just keep the iframe as a part of this repo. If there is a bug in the iframe code would be equivalent to if there is a bug in the library and would be handled as any bug in the lib. I don't see any major benefits to separating them, but it adds a bunch of complexity in our development process.

oed:
Actually I think I've changed my mind on 2. If we are interested in storing keys in the iframe in the future (which is required if we want to support users w/o an eth-wallet) we need to be able to upgrade the iframe without loosing users keys. I guess there might be ways around that, but not with good UX.

zachferland:
hmm yeah had not thought about that, in that case I just thinks its important that its on our domain. Also doesn't really matter where we put it for our first iteration, change is low cost right now, any change would just be a little time extra for full sync on first use again.

zachferland:
Ready for a review (cc @oed @c-castillo ) Just has minor remaining tasks, but won't change the code much:

- [ ] reference iframe once released
- [ ] up Orbitdb version, once pr there is released, but we can also move ahead without for now, knowing that it will go in soon.

Enhancements/Checks
- [ ] tests sometimes failing on CI, but maybe just improve reliability in another PR
- [ ] there may be ways to improve iframe loading times 
- [ ] tests in some further examples once iframe is deployed 

Also note for myself, check that ipfs proxy waits for ipfs to load in iframe before returning response to first message

example/index.js
```diff
@@ -1,10 +1,28 @@
...
+  }
+
+  /**
+   *  This uses a local ipfs-js instance and local orbitdb cache instead of the
+   *  shared iframe storage available at 3box.io. Easier for testing and differing
+   *  configs. But in production you will get the best performance by using the
+   *  default iframe configuration.
+   */
+  const ipfs = new window.Ipfs(IPFS_OPTIONS)
+  const opts = { ipfs, iframeStore: false }
```
oed: I'm a bit worried about having this in the example as people will look at it and most likely copy from the example.

oed: Maybe we can have two different examples?

package.json
```diff
@@ -44,10 +44,13 @@
     "bip39": "^2.5.0",
     "ipfs": "^0.32.2",
     "ipfs-api": "^24.0.0",
+    "ipfs-postmsg-proxy": "^3.1.1",
     "ipfs-pubsub-room": "^1.4.0",
     "js-sha256": "^0.9.0",
     "muport-core": "^0.4.1",
     "orbit-db": "^0.19.9",
+    "orbit-db-cache": "^0.2.4",
```
oed: Why do we need this? Looking at the code in 3box.js we could just use the default in orbit by not passing an alternative cache?

zachferland: ah yup, good catch, this was before I made the pr to orbitdb and I had different implementation

example/index.js
zachferland: im fine just using the iframe by default, this was just before we had it deployed anywhere.

zachferland:
yeah 204 change has to be added, its not a shared config though, it is just moved now to the iframe

zachferland:
although on second thought, that ipfs config should be exported somewhere now that users may pass there own ipfs object, that won't be used much though

package.json
oed: ok :)
Unrelated but I'm pretty sure we can remove the `ipfs-pubsub-room` dep.

example/index.js
oed: ah, I see!

PR #198, opened 2018-12-03
Now published did to public profile if not there already

oed:
This PR adds code to publish the users DID to the public profile. This is makes it easier to verify users in various ways.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -101,6 +101,7 @@ class Box {
         }
         if (syncPromises.length === 2) {
           await Promise.all(syncPromises)
+          await this._ensureDIDPublished()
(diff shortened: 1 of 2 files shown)


oed:
@zachferland hm, that's a good point. We can just do it at creation now as well maybe? Going to take a closer look if that's a good idea tomorrow!

oed:
So I'm unsure if we'll be able to do this deterministically. Seems like if we want to add it as the first entry we can't know if there has been a quick replication from the network yet. So the first time it would be added in the beginning, but other times it might be added later. The way it's done now seems like a more reliable way of ensuring that the did is only added once to the log.
Merging this for now.

PR #201, opened 2018-12-03
docs: add note on how to load from cdn

zachferland:


README.md
```diff
@@ -25,11 +25,25 @@ Import the 3box module
...
+```html
+<!-- The most recent version  -->
+<script src="https://unpkg.com/3box/dist/3box.js"></script>
+<!-- The most recent minified version  -->
+<script src="https://unpkg.com/3box/dist/3box.min.js"></script>
+<!-- Load specific versions by specifying the version as follows -->
+<script src="https://unpkg.com/3box@<version>/dist/3box.js"></script>
+```
+
+You can also load it using a <script> using the unpkg CDN or the jsDelivr CDN. Inserting one of the following lines will make a Ipfs object available in the global namespace.
```
oed: You mean the Box object and not the Ipfs object right? 😂

zachferland: ahh yeah, don't want any of that line :) pushed up a fix

Issue #197, opened 2018-11-30
Implement verifier of accouts for profile objects.


oed:
Base work on the https://github.com/3box/3box-js/tree/feat/verified-accounts branch.

Add the functionality to check the verification of accounts in any users profile. 

1. In the file src/3box.js implement the `getVerifiedAccounts` function. It takes one parameter, a profile object (that you can get from the `getProfile` method for any user). It then checks the `'github'`, and `'twitter'` keys for proof urls. If they are present it validates the proof (using the functions in `src/utils/verifier.js`). It then returns an object that looks like this:
```js
{
  github: "oed",
  twitter: "oedbeo"
}
```
With the given users proved handles for each account.

Note that this is blocked by #195 & #196

Issue #195, opened 2018-11-30
Implement github verification


oed:
Base work on the https://github.com/3box/3box-js/tree/feat/verified-accounts branch. 

Add the functionality to verify a github account to 3box-js. This is done by posting the users DID (which should be in the public profile) to a gist file on github. There are skeleton files provided which also include jsdoc that describes the functionality of the functions to implement. 
1. In the file `src/utils/verifier.js` implement the `verifyGithub` function. It takes two parameters, did & gistUrl. This method should simply check that the did is present in the gist, then return the github handle of the user.
2. In the file `src/verifications.js` implement the `github` and `addGithub` methods by using the previously implemented method. In the `addGithub` method, if the gistUrl is valid it should add this url to the `'github'` key in the public profile. The `github` method can then get the url from this entry.

PR #209, opened 2018-12-06
Support the verification of github accounts

joaoaguiam:
Based on the discussions on issue #334. Solving issues #195 and #197


src/utils/verifier.js
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Verifies that the gist contains the given muportDID and returns the users github handle.
+   * Throws an error otherwise.
+   *
+   * @param     {String}            did                     The muport DID of the user
+   * @param     {Object}            gistUrl                 URL of the proof
+   * @return    {String}                                    The github handle of the user
+   */
+  verifyGithub: async (did, gistUrl) => {
+    console.log({did, gistUrl});
```
oed: Let's remove this 📝

src/utils/verifier.js
```diff
@@ -0,0 +1,39 @@
...
+    if (!gistUrl || gistUrl.trim() === "") {
+      throw new Error("The proof of your Github is not available");
+    }
+    if (!did || did.trim() === "") {
+      throw new Error("DID parameter not provided");
+    }
+    let gistFileContent = await (await fetch(gistUrl)).text();
```
oed: Please use httpRequest (imported above) for this. The main reason is to be consistent with how we do http requests.

src/utils/verifier.js
```diff
@@ -0,0 +1,39 @@
...
+    if (gistFileContent.trim() !== did) {
+      throw new Error("Gist File provided does not contain the correct DID of the user");
+    }
+
+    const githubUsername = gistUrl.replace("https://gist.githubusercontent.com/", "").split("/")[0];
```
oed: Why do you need to do `.replace` here? This should do the trick:
```suggestion
    const githubUsername = gistUrl.split("/")[3]
```

src/verifications.js
```diff
@@ -0,0 +1,113 @@
...
+   * If verification is successful, proof is stored in the public store.
+   * Throws an error if verification is not successful.
+   *
+   * @param     {Sting}               key                     Account key to be stored in the public profile
+   * @param     {String/Object}       proof                   This param will be sent to the verification function and stored in the 3box
+   *                                                          profile for future verification
+   * @param     {Function}            verificationFunction    Function receiving the user DID and the proof received as param.
+   *                                                          This function should return the username to be stored in the 3box profile
+   */
+  async _addVerifiedPublicAccount(key, proof, verificationFunction) {
```
oed: Like these abstractions 😸

src/verifications.js
```diff
@@ -0,0 +1,113 @@
...
+   * @return    {String}                                    The twitter handle of the user
+   */
+  async addTwitter (tweetUrl) {
+    return false
+  }
+
+  /**
+   * Returns all the verified accounts 
+   */
+  async getVerifiedAccounts () {
```
oed: Don't think we need this function. In most dapps you'd want to run `.twitter()` and `.github()` etc separately to make the dapp load more quickly (you get the results when each finish rather than having to wait for all of the to finish). Let's remove this function for now.

Btw, this doesn't solve #197 which should be a static function in the `Box` class to which you pass a profile object. It's meant as a way of verifying accounts of other users. 
You don't need to address that in this PR though. Sorry for the confusion 🙂 Let me know if anything is unclear!

src/3box.js
```diff
@@ -199,6 +204,15 @@ class Box {
     }
   }
 
+  /**
+   * Verifies the proofs of social accounts that is present in the profile.
+   *
+   * @param     {Object}            profile                 URL of the Address Server
+   * @return    {Object}                                    An object containing the accounts that have been verified
+   */
+  static async getVerifiedAccounts (profile) {
```
oed: This is the skeleton code for #197, but disregard for now!

src/verifications.js
```diff
@@ -0,0 +1,113 @@
...
+   * Internal method to retrieve the verified value for a given key. It will verifiy if the proof is still valid
+   * 
+   * @param {sting}     key - Account key to be retireved from the public profile
+   * @param {function}  verificationFunction - Function receiving the user DID and the proof received as param.
+   * This function should return the username of the user that will be compared to the value stored  in the 3box profile.
+   */
+  async _getVerifiedPublicAccount(key, verificationFunction) {
+    this._requireBox();
+    const proof = await this._box.public.get("proof_" + key);
+    console.log(proof);
```
oed: Remove this line 
```suggestion
```

PR #211, opened 2018-12-10
 graphql and rest api/cache

zachferland:


zachferland:
This is the roughly the code now, but a few remaining tasks:

Remaining tasks:

- [ ] Rewire some old test
- [ ] Add new test for api calls and graphql 
- [ ] Doc clean up 
- [ ] Will have to rebase, and rebase on other open PRs or vice versa (touch some of the same code)

src/__tests__/3box.test.js
```diff
@@ -114,7 +125,8 @@ describe('3Box', () => {
     const IPFS_OPTIONS = {
       EXPERIMENTAL: {
         pubsub: true
-      }
+      },
+      repo: './.js-ipfs'
```
oed: Why this path rather than the ./tmp/ipfs<n> that we have been using so far?

src/3box.js
```diff
@@ -13,10 +14,14 @@ const utils = require('./utils')
 
 const ADDRESS_SERVER_URL = 'https://beta.3box.io/address-server'
 const PINNING_NODE = '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmZvxEpiVNjmNbEKyQGvFzAY1BwmGuuvdUTmcTstQPhyVC'
+// const PINNING_NODE = '/ip4/127.0.0.1/tcp/4003/ws/ipfs/QmcHUWf1YemBYrezctp6cSyZpSAANz2deFb9As3EMZnuXf'
```
oed: remove this line

src/3box.js
```diff
@@ -13,10 +14,14 @@ const utils = require('./utils')
...
 const PINNING_ROOM = '3box-pinning'
 const IFRAME_STORE_VERSION = '0.0.2'
 const IFRAME_STORE_URL = `https://iframe.3box.io/${IFRAME_STORE_VERSION}/iframe.html`
 
+const GRAPHQL_SERVER_URL = 'https://aic67onptg.execute-api.us-west-2.amazonaws.com/develop/graphql'
```
oed: I'm guessing we change this once cristobal is back and can set up a proper url?

src/3box.js
```diff
@@ -75,6 +80,9 @@ class Box {
     this._rootStore = await this._orbitdb.feed(rootStoreName)
     const rootStoreAddress = this._rootStore.address.toString()
 
+    console.log('rootStoreAddress')
+    console.log(rootStoreAddress)
```
oed: These shouldn't be here.

src/3box.js
```diff
@@ -143,9 +151,40 @@ class Box {
    * @param     {Object}    opts.ipfs               A js-ipfs ipfs object
    * @param     {String}    opts.orbitPath          A custom path for orbitdb storage
    * @param     {Boolean}   opts.iframeStore        Use iframe for storage, allows shared store across domains. Default true when run in browser.
+   * @param     {Boolean}   opts.api                Use 3Box API to fetch profile instead of OrbitDB. Default true.
```
oed: Maybe this param is more clear if it's called something like `useCache`?

src/3box.js
```diff
@@ -392,4 +446,18 @@ async function getRootStoreAddress (serverUrl, identifier) {
   })
 }
 
+async function getProfileAPI (rootStoreAddress, serverUrl) {
```
oed: Should this also start with `_`?

src/3box.js
```diff
@@ -211,6 +250,21 @@ class Box {
     }
   }
 
+  /**
+   * GraphQL for 3Box profile API
+   *
+   * @param     {Object}    query               A graphQL query object.
+   * @param     {Object}    opts                Optional parameters
+   * @param     {String}    opts.graphql        URL of graphQL 3Box profile service
```
oed: Maybe something like `opts.serviceUrl` is more descriptive?

oed: Maybe `graphqlServer` to be consistent.

src/__tests__/3box.test.js
zachferland: testsweren't using ' ./tmp/ipfs' at the moment, maybe that was lost somewhere else in another change, can check, but yeah lets use tmp

src/3box.js
zachferland: yup thanks

src/3box.js
zachferland: 👍

src/3box.js
zachferland: yeah graphqlServer is good

src/3box.js
zachferland: 👍

src/3box.js
zachferland: hmmm could be, feels like cache is less descriptive of what it is, since not clear what kind of cache? what its rules? could it be out of date? when its really just a http api call that happens to also have a cache. Cache feels like it would be in support of orbitdb as well instead of just around it. But idk `api` alone is not super useful. But zooming out, if we consider that api overall a cache of orbitdb, which I guess we have, then that wording would usefully indicate what its purpose is.

oed: Idk, I feel like api is way to generic. Is `useCacheService` better?

zachferland: yep that works for me, lets do 'useCacheService', I think service is a good distinction

src/3box.js
zachferland: yeah probably, may get to it, but not highest priority and Cristobal can do it more quickly

PR #227, opened 2018-12-14
Store link consent (proof for ethereum address) in public store

oed:
Didn't add an explicity function to make the profile public. Instead I stored the proof for the ethereum address in the public store. This should reduce the number of times the linkProfile consent message is shown to only once (before you'd have to do it every time you cleared your cache).


README.md
```diff
@@ -1,9 +1,10 @@
-[![CircleCI](https://img.shields.io/circleci/project/github/3box/3box-js/master.svg?style=for-the-badge)](https://circleci.com/gh/3box/3box-js)
+[![CircleCI](https://img.shields.io/circleci/project/github/3box/3box-js.svg?style=for-the-badge)](https://circleci.com/gh/3box/3box-js)
```
zachferland: removed specific branch? want develop or master? and would need same change in readme template

oed: I'm pretty sure I just generated from the template. It shouldn't be master, and I think not specifying gives the default one.

zachferland: ah yeah I see now in template, yeah not specifying is develop, 👍

PR #230, opened 2018-12-15
feat: use fetch

oed:
This should be simple enough. Please test locally with example for various scenarios of how the address-server is used.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,4 +1,4 @@
-const XMLHttpRequest = (typeof window !== 'undefined') ? window.XMLHttpRequest : require('xmlhttprequest').XMLHttpRequest
+const fetch = (typeof window !== 'undefined') ? window.fetch : require('node-fetch')
 const Multihash = require('multihashes')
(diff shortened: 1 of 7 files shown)


src/utils/index.js
```diff
@@ -46,34 +46,18 @@ module.exports = {
...
-        }
-      }
-      request.open(method, url)
-      request.setRequestHeader('accept', 'application/json')
-      // request.setRequestHeader('accept', '*/*')
-      if (method === 'POST') {
-        request.setRequestHeader('Content-Type', `application/json`)
-        request.send(JSON.stringify(payload))
-      } else {
-        request.setRequestHeader('Authorization', `Bearer ${payload}`)
```
zachferland: Token not passed in header anymore?

oed: Yeah, I'm not sure why we did that. Don't think the address-server checks that, but we can double check.

oed:
Do you think we should switch to cross-fetch in this PR or hold of for now?

zachferland:
nah, not necessary in this pr

PR #233, opened 2018-12-20
feat: add twitter verifier

oed:
This could should work, but needs to be tested when the serverless twitter verification function is written.

TODO:
- [x] add tests
- [x] add https-did-resolver


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "./node_modules/.bin/standard --verbose src/**",
-    "test": "rm -rf ./tmp ; jest --forceExit --detectOpenHandles --coverage --runInBand --testURL=\"http://localhost\"",
```
c-castillo: why are you removing `--forceExit`? running test fails because of a timeout, at least at the CI. We can create a special method for the CI though

oed: Yeah, just noticed this. I removed it to get the console output locally. Readded now.

PR #235, opened 2018-12-20
fix: wait for iframe to load, and up orbit proxy cache with better su…

zachferland:
…pport

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -292,9 +292,6 @@ class Box {
    * @return    {Box}                                       the 3Box instance for the given address
    */
   static async openBox (address, ethereumProvider, opts) {
-    if (iframeLoadedPromise) {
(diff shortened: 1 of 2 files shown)


src/3box.js
```diff
@@ -287,6 +292,9 @@ class Box {
    * @return    {Box}                                       the 3Box instance for the given address
    */
   static async openBox (address, ethereumProvider, opts) {
+    if (iframeLoadedPromise) {
+      await iframeLoadedPromise
```
oed: This is needed in the `getProfile` using orbitdb also right?

oed:
@zachferland I moved the await for the iframe promise to initIPFS. Seemed more logical to keep it there.

zachferland:
great, yeah thats a good spot for it 👍

PR #239, opened 2019-01-02
docs: how to change ipfs configs

zachferland:


zachferland:
Docs to resolve - https://github.com/3box/3box-js-iframe/issues/2

PR #243, opened 2019-01-07
feat: implement pinning-node check and reconnect

oed:
Now in every call to `set` & `remove` we check if we are connected to the pinning node. If we are not we reconnect to make sure that all updates gets synced to the pinning node.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -68,49 +68,33 @@ class Box {
     const didFingerprint = utils.sha256Multihash(did)
     const rootStoreName = didFingerprint + '.root'
 
-    const pinningNode = opts.pinningNode || PINNING_NODE
-    // console.time('start ipfs')
+    this.pinningNode = opts.pinningNode || PINNING_NODE
     this._ipfs = await initIPFS(opts.ipfs, opts.iframeStore)
-    // console.timeEnd('start ipfs')
-    // TODO - if connection to this peer is lost we should try to reconnect
-    // console.time('connect to pinning ipfs node')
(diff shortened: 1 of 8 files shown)


oed:
Hm, seems like the test fail because the ipfs in the testfile connects to a bunch of random ipfs nodes.

src/3box.js
```diff
@@ -378,6 +346,15 @@ class Box {
     }
   }
 
+  async _ensurePinningNodeConnected (odbAddress) {
+    const roomPeers = await this._ipfs.pubsub.peers(odbAddress)
+    if (!roomPeers.find(p => p === this.pinningNode.split('/').pop())) {
+      this._ipfs.swarm.connect(this.pinningNode)
+      const rootStoreAddress = this._rootStore.address.toString()
+      this._pubsub.publish(PINNING_ROOM, { type: 'PIN_DB', odbAddress: rootStoreAddress })
```
zachferland: Does this not have to wait for the peer (pinning node to be connected) before pushing this message? And should any of this block/await before setting or removing?

zachferland: oh well i guess it doesnt have to block set or remove, since it will eventually share those updates once connected

oed: Yeah, there are two cases:
1. We lost the connection to the pinning node. Then the onPeer function will send the msg to the pinning node once connected.
2. We are connected to the pinning node, but it doesn't have our db open. Then we just publish to the pinning room.

oed: Initially I had separate logic for bot cases, but that was a bit redundant I think.

PR #250, opened 2019-01-10
Release/1.2.0

oed:


src/3box.js
```diff
@@ -202,13 +202,13 @@ class Box {
       ipfs = globalIPFS
       usingGlobalIPFS = true
     } else {
-      ipfs = await initIPFS(opts.ipfs, opts.iframeStore)
+      ipfs = await initIPFS(opts.ipfs, opts.iframeStore, opts.ipfsOptions)
```
c-castillo: `iframeStore` will be undefined, not sure if that can generate an error

oed: nope, that's fine (see the `initIPFS` code)

PR #251, opened 2019-01-10
Fixes #248

### LINKED ISSUE #248: Calling getprofile returns the error "utils.fetchJson is not a function"
**Describe the bug**
A simple `await Box.getProfile('0x0c269b2d1376b2a2a5222564f9c8ea0065dd2c79');` will return an error saying `utils.fetchJson is not a function`

Bug is quite simple : npm package contains a `utils.js` file along a `/utils` folder. Webpack will use the file when importing module, but won't find the `fetchJson()` function.
The `utils.js` file should not be there.

**To Reproduce**
1. Install package 3box `v1.1.0`
2. Try to get a profile with `Box.getProfile()` method
4. Error pops-in console

sethfork:
Fixes #248 "Calling getprofile returns the error "utils.fetchJson is not a function""

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules/
 lib/
-dist/
 coverage/
 tmp/
 .js-ipfs
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "3box",
   "version": "1.1.0",
   "description": "Interact with user data",
-  "main": "lib/3box.js",
+  "main": "dist/3box.js",
   "directories": {
     "lib": "lib"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "3box",
-  "version": "1.0.6",
+  "version": "1.1.0",
   "lockfileVersion": 1,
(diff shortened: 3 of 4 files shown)


oed:
There should be a fix in version 1.2.0. Can you confirm? Will close this for now, but I'll give you some feedback on it regardless :)

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "3box",
   "version": "1.1.0",
   "description": "Interact with user data",
-  "main": "lib/3box.js",
+  "main": "dist/3box.js",
```
oed: The js fine in the dist directory is a file compiled with all the dependencies into one single file. Using it as the main file is not so good for people that have their own build processes set up to optimize all dependencies since 3box will only be a blob. the `lib` folder contains just the 3box code (compiled down to ES5) without the dependencies baked in.

.gitignore
```diff
@@ -1,6 +1,5 @@
 node_modules/
 lib/
-dist/
```
oed: Since the files in the `dist` folder contain all dependencies baked into one file it's not really supposed to got into git.

PR #270, opened 2019-01-18
Feat/3id module

oed:
This moves all keymanagement logic into a separate module. It will help a lot for Spaces.


oed:
Thanks, yeah it untangles stuff quite a bit. There is still some small improvements that can be made here and there though. I have created stories for some of the things, but don't have an extensive list. 
Think this should be fine to merge for now!

PR #272, opened 2019-01-21
Feat/create spaces class

oed:
To be merged after #270 

This implements the spaces class. 
I'm not completely convinced that `box.space.nameOfSpace` is the best api. It might confuse developers since `box.space` is also a class. Suggestions for how to improve it is welcome :)


src/keyValueStore.js
```diff
@@ -81,7 +82,11 @@ class KeyValueStore {
   }
 
   async _load (odbAddress) {
-    this._db = await this._orbitdb.keyvalue(odbAddress || this._name)
+    const key = this._3id.getKeyringBySpaceName(this._name).getDBKey()
+    this._db = await this._orbitdb.keyvalue(odbAddress || this._name, {
+      key,
+      write: [key.getPublic('hex')]
+    })
```
zachferland: is this adding the key as a signing key to store and giving it write permissions?

zachferland: vs init orbitdb took a keystore adapter before and did it for each store we initialized

src/spaces.js
```diff
@@ -0,0 +1,32 @@
...
+    if (!entries.find(entry => entry.payload.value.odbAddress.split('.')[2] === name)) {
+      this._rootStore.add({ odbAddress: spaceAddress })
+    }
+    const syncSpace = async () => {
+      await this[name]._sync()
+      if (opts.onSyncDone) opts.onSyncDone()
+    }
+    syncSpace()
+    return this[name]
+  }
```
zachferland: 👍

oed:
Rebased this on develop

src/keyValueStore.js
oed: So the orbitdb keystore could only retreive keys by ipfs node id. So we could only have one key. This key was added with write permissions by default to all stores that where created without a `write` array. Since with spaces we need to have different keys for each space we need to add the signing key and set the write permissions.

oed: So yes, you are right 🙂

oed:
> We should also consider if the space names are keyed or not, since it makes the error handling less clear. For example i try box.spaces.zachdapp.public it will just say can't get public on undefined, vs if was box.space('zachdapp') it could throw an error saying you must open space first or similar error handling. (although i do like the simplicity of the ui for the keyed space)

Yeah that's a good point. Another thing is that if we have a `.` in the space name the api will break as well.

oed:
How about this, mostly taken from your suggestions:

`box.openSpace('name')` - to open space, returns the space
`box.getSpace('name')` - returns the space if opened
`box.spaces` - an object that contains all spaces with names as keys

This should cover all use cases right?

zachferland:
yeah that works, cant see any reason not to offer both `box.getSpace` and `box.spaces` as mentioned. Maybe `getSpace` ends up being used more, maybe not, but open spaces has to be saved on object somewhere, so makes sense to have `box.spaces`

src/keyValueStore.js
zachferland: ah yup, great, thanks for the extra details!

oed:
Updated the PR with the new api. Decided to not include `getSpace` since we can have the same functionality if you call `openSpace` on a space that has already been opened.

PR #290, opened 2019-01-25
fix: resolve sync promise if there are more entries locally than on p…

oed:
…inning node

This should fix infinite onSyncDone!


oed:
I think this should resolve atleast some of the issues where onSyncDone wasn't getting called.

oed:
This is already in both master and develop

PR #296, opened 2019-01-28
Feat/refac configs

zachferland:
This pr does the following:

- Move configs to single place 
- add env variables as an options
- allow those env variables to only be used in a dev build, all other build ignore env var and use defaults (primarily so that we don't accidentally publish with different env vars)
- creates a dev build that allows above, and runs in watch mode (auto rebuild)
- allows muport ipfs configs to be passed through


oed:
Oh, great! I've wanted something like this for a while 😄

c-castillo:
Awesome! @zachferland I also wanted to do this a long time ago

zachferland:
This is all good now for review and merge! Checked all the builds to confirm, env var work as described above.

PR #302, opened 2019-01-30
fix: close ipfs, so reopen works

zachferland:
This fixes, where you openBox, close and then try to open again, now onsyncdone is called both times. 
This was the minimal code fix but after looking into it more it looks like ipfs/pubsub with multiple instances causes some issues, I think @oed  has mentioned this before. So there could be some issues around caching in these instances that cause pubsub to not send any messages and for onsyncdone to not be returned. I might makes sense to generally use the global ipfs/orbitdb and not close them, just keep open and then connect if already available. 

Seem likes ipfs was caching something, since peers were listed before onNewPeer was even called once. And then anytime you tried to create a new pubSub (after one was already created on an instance) it would look like its available but then just would not publish any messages.

Also if you create an ipfs instance, call stop on it, then create a new instance and call stop again it throws an error saying stop was already called. So need to understand what ipfs is doing, seems to keep a singleton instance around or something. Its probably best to treat it as a node/service running in the background instead of we do now.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -82,12 +82,12 @@ class Box {
     const rootStoreName = this._3id.muportFingerprint + '.root'
 
     this.pinningNode = opts.pinningNode || PINNING_NODE
-    this._ipfs = await initIPFS(opts.ipfs, opts.iframeStore, opts.ipfsOptions)
+    this._ipfs = globalIPFS || await initIPFS(opts.ipfs, opts.iframeStore, opts.ipfsOptions)
+    globalIPFS = this._ipfs
     this._ipfs.swarm.connect(this.pinningNode, () => {})
 
     // const cache = (opts.iframeStore && !!cacheProxy) ? cacheProxy : null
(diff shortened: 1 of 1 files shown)


oed:
Ugh, the reason it was commented out is because it breaks the tests :/

zachferland:
hahah yeah, something is up with, because like I mentioned after the first time you call, not even on the "same" instance it keeps saying stop was already called. 

I think the solution is to just keep IPFS/orbit in the background, and change the pin db logic a bit, so that it sends on new peer the first, and after just sends when opening again if already connect to the peer

oed:
Ok, yeah that seems alright

zachferland:
This keeps ipfs around and doesn't stop it. Just reuses it if available. More stable than before, fixes much of the weird behavior in stopping/starting and with pubsub.

One thing not clear, is why opening box the second time after closing is much slower than the first time. Something to look into in the future.

zachferland:
@oed  all set now, rewrote over other commits, since the approach completely changed, pretty minimal here

PR #305, opened 2019-01-31
feat: onsync done for spaces

zachferland:
Implemented with minimal code diff, but this was also a good area for refactor. We should have two additional logical services, one that handles pubsub messages, on peers joining, and more importantly messages on response now that it is not only relevant to openbox, but also spaces, and they happen at differing times. Second we could probably have a logical service that handles syncing state across dbs/spaces. 

In the current implementation the pinDB message to the pinning server results in the pinning server opening all stores, including all spaces. We will likely want to change to this to open what is needed when it is needed.

We could also refactor some the code around open/load/sync as similar handling is need for both primary DBs and spaces, but code exist in two different places at the moment.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -131,8 +131,7 @@ class Box {
           hasResponse[pubStoreAddress] = true
         }
         if (spaceMessageFilterActive && data.odbAddress.includes('space') === true) {
-          const spaceName = data.odbAddress.split('/')[3].split('.')[2]
-          this.spacesPubSubMessages[spaceName] = data
+          this.spacesPubSubMessages[data.odbAddress] = data
         }
         if (syncPromises.length === 2) {
           const promises = syncPromises
@@ -347,8 +346,7 @@ class Box {
     if (!this.spaces[name]) {
       this.spaces[name] = new Space(name, this._3id, this._orbitdb, this._rootStore, this._ensurePinningNodeConnected.bind(this))
       try {
-        const entryMessage = this.spacesPubSubMessages[name]
-        opts = entryMessage ? Object.assign({ numEntries: entryMessage.numEntries }, opts) : opts
+        opts = Object.assign({ numEntriesMessages: this.spacesPubSubMessages }, opts)
         await this.spaces[name].open(opts)
       } catch (e) {
         console.log(e)
--- a/src/3box.js
+++ b/src/3box.js
@@ -111,7 +111,16 @@ class Box {
     let syncPromises = []
     let hasResponse = {}
 
+    // Filters and store space related messages for 3secs, the best effort
+    // simple approach, until refactor
+    let spaceMessageFilterActive = true
+    let filterTimeSet = false
+
     const onMessageRes = async (topic, data) => {
+      if (!filterTimeSet) {
+        filterTimeSet = true
+        setTimeout(() => { spaceMessageFilterActive = false }, 3000)
+      }
       if (data.type === 'HAS_ENTRIES') {
         if (data.odbAddress === privStoreAddress && !hasResponse[privStoreAddress]) {
           syncPromises.push(this.private._sync(data.numEntries))
@@ -121,7 +130,7 @@ class Box {
           syncPromises.push(this.public._sync(data.numEntries))
           hasResponse[pubStoreAddress] = true
         }
-        if (data.odbAddress.includes('space') === true) {
+        if (spaceMessageFilterActive && data.odbAddress.includes('space') === true) {
           const spaceName = data.odbAddress.split('/')[3].split('.')[2]
           this.spacesPubSubMessages[spaceName] = data
         }
--- a/src/3box.js
+++ b/src/3box.js
@@ -68,6 +68,9 @@ class Box {
      * @property {Object} spaces            an object containing all open spaces indexed by their name.
      */
     this.spaces = {}
+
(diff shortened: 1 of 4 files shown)


src/3box.js
```diff
@@ -76,6 +76,10 @@ class Box {
      * @property {Object} spaces            an object containing all open spaces indexed by their name.
      */
     this.spaces = {}
+    /**
+     * @property {Object} spacePubSubMessages   local store of all pinning server pubsub messages seen related to spaces
+     */
```
oed: This will become part of the API documentation. I don't think this should be there.

src/3box.js
```diff
@@ -125,6 +129,11 @@ class Box {
           syncPromises.push(this.public._sync(data.numEntries))
           hasResponse[pubStoreAddress] = true
         }
+        if (data.odbAddress.includes('space') === true) {
```
oed: 😂 these are always fun.

Also, remove the console log on the next line.

src/3box.js
```diff
@@ -342,6 +351,12 @@ class Box {
       } catch (e) {
         delete this.spaces[name]
       }
+      const syncSpace = async () => {
+        const numEntries = this.spacesPubSubMessages[name] ? this.spacesPubSubMessages[name].numEntries : null
```
oed: Wouldn't it be easier to send the number of msgs to the open function rather than to move code into the 3box main module?

src/3box.js
zachferland: right, yeah wanted note for myself, but shouldn't be js doc

src/3box.js
zachferland: 👍

src/3box.js
zachferland: oh yeah, if this is a minimal code change for now, yeah I should keep this over in the open function  👍

zachferland:
> Could there be an issue if you open the space right after you do openBox and the number has not been saved yet?

right now its simplistic, if not available it falls back to the 3 second wait. Ideally if it didnt have the message yet it would start wait, but would resolve early if it did get the message before 3 seconds. But thats a larger change, i think it makes more sense as part of a refactor story, as described above.

src/3box.js
oed: I guess you can add @private then

src/3box.js
oed: Oh, haha! I didn't think of that either :P

src/keyValueStore.js
```diff
@@ -52,7 +52,8 @@ class KeyValueStore {
   async _sync (numRemoteEntries) {
     this._requireLoad()
     // let toid = null
-    if (numRemoteEntries <= this._db._oplog.values.length) return Promise.resolve()
+    const numEntriesDefined = !(numRemoteEntries === null || numRemoteEntries === undefined)
+    if (numEntriesDefined && numRemoteEntries <= this._db._oplog.values.length) return Promise.resolve()
```
oed: Why is `numEntriesDefined` needed? couldn't the if statement simply be this?
```js
if (numRemoteEntries && numRemoteEntries <= this._db._oplog.values.length) return Promise.resolve()
```

src/space.js
```diff
@@ -24,10 +24,7 @@ class Space {
         this._rootStore.add({ odbAddress: spaceAddress })
       }
       const syncSpace = async () => {
-        // TODO - this logic isn't completely sound yet. Now it will just
-        // always resolve after three seconds. We need a way to get unsynced
-        // entries for the given store from the pinning node.
-        await this._store._sync()
+        await this._store._sync(opts.numEntries || null)
```
oed: I guess it's fine to just send undefined here, so `|| null` shouldn't be needed.

src/keyValueStore.js
zachferland: problems is 0 vs undefined vs null in js - some examples 
null < 1  = true
undefined < 1 = false 
1 && 1 < 2 = true 
0 && 0 < 2 = 0  (which is falsey)

yeah caused some trouble for a little while, so just made it explicit

oed: ah, I see! Annoying :P

zachferland:
All set for final review, all though not optimal this keeps it simple, until we have a more thoughtful refactor, as there will be a few better ways to deal with it there. 

The simplistic approach here will store all space messages it sees, so the simple fix for now, is it only listens for messages for 3 secs, where it will likely fetch relevant ones, and may get other messages, but not so many that storing them is any issue.

src/3box.js
```diff
@@ -118,6 +130,10 @@ class Box {
           syncPromises.push(this.public._sync(data.numEntries))
           hasResponse[pubStoreAddress] = true
         }
+        if (spaceMessageFilterActive && data.odbAddress.includes('space') === true) {
+          const spaceName = data.odbAddress.split('/')[3].split('.')[2]
+          this.spacesPubSubMessages[spaceName] = data
```
oed: It seems like this will just save whatever space with the given name is called last. So if box A & B both open space `S` there could be a problem where A thinks it has the same number of entries as B.
To solve this we likely need to store the entire odbAddress and send spacesPubSubMsgs along to the `open` method.

zachferland: ah yes yes, constantly forgetting the message the space is a global space, nice catch, yeah i think the sol you mentioned is the simplest at the moment

PR #312, opened 2019-02-01
ref(3id): removes muport-core dependancy

oed:


zachferland:
Not sure what happen here, has code from other branches in one commit. At first glance, just the 3id part makes sense, will review more in depth once just these changes here. So it is still technically the same form of a muport did?

oed:
oh, I must have messed up when I rebased against develop. I'll push a fix!

oed:
I did `git commit -a --amend` instead of `git rebase --continue` 🙄

zachferland:
ah yup, great, thanks! will review this weekend

oed:
Also did an update to muport-did-resovler so that it only fetches the ipfs document for now and does not look at the blockchain. This reduced the build size quite a bit. Mostly because we don't use `ethr-did-resolver` anymore.
https://bundlephobia.com/result?p=3box@1.3.0-beta.2

src/3id/index.js
```diff
@@ -10,49 +10,60 @@ const STORAGE_KEY = 'serialized3id_'
...
-    this.muportFingerprint = utils.sha256Multihash(this._muport.getDid())
+  }
+
+  async _initMuport (muportIpfs) {
+    let keys = this._mainKeyring.getPublicKeys()
+    const doc = createMuportDocument(keys.signingKey, this.managementAddress, keys.asymEncryptionKey)
+    let docHash = (await this._ipfs.files.add(Buffer.from(JSON.stringify(doc))))[0].hash
+    this._muportDID = 'did:muport:' + docHash
+    this.muportFingerprint = utils.sha256Multihash(this._muportDID)
+    const publishToInfura = async () => {
```
zachferland: why does this use both this._ipfs and ipfsMini?

zachferland: oh yes, one is adding locally, and the other publish to infura as made clear by the function name :)

src/3box.js
```diff
@@ -47,9 +47,10 @@ class Box {
   /**
    * Please use the **openBox** method to instantiate a 3Box
    */
-  constructor (threeId, ethereumProvider, opts = {}) {
+  constructor (threeId, ethereumProvider, ipfs, opts = {}) {
     this._3id = threeId
     this._web3provider = ethereumProvider
+    this._ipfs = ipfs
```
zachferland: 👍

src/3id/index.js
oed: yeah, ideally we can remove the infura one in the future :)

PR #313, opened 2019-02-04
feat: es6 modules, api call refactor, separate api only small build

zachferland:
Need some more time to review parts. This is bound to break something for someone. Considering doing another branch with each part in separate commits. Also considering only adding the api refactor and api read only lib. Then es6 modules at a later point, may cause too much trouble at the moment.

The separate api refactor is definitely worth shipping as soon as possible though. Gives a nice 80kb lib for anyone who only wants to get profiles from out api.


zachferland:
Actually the best short term thing would be to just keep commonjs for babel but write all our stuff in es6, so that we can get more the build benefits at later point, but minimize the chance of breaking anything right as this moment.

zachferland:
Now split into #316  and #315 so that we can ship them separately, or most importantly just #315. Closing this one now.

At this moment switching to es6 only save 100mb on the build.

PR #314, opened 2019-02-04
Add email to box.verified API

c-castillo:


src/utils/verifier.js
```diff
@@ -53,6 +53,31 @@ module.exports = {
...
+    if (verified.payload.sub !== did) {
+      throw new Error('Verification not valid for given user')
+    }
+    const claimData = verified.payload.claim
+    if (!claimData.email_address) {
+      throw new Error('The claim for your email address is not correct')
+    }
+    return {
+      email_address: claimData.email_address,
+      proof: claimData.code,
```
oed: We don't really need this code here. Not sure why it's included in the response claim anyway?

c-castillo: @oed I agree it feels forced to add the code as a proof, but since pair `proof_email, proof` is stored in the user's 3Box, I thought it was the best to add some proof rather than a `True` or `null` value.

oed: I'm not sure I follow. `proof_email` is simply the claim from the verification service. This return function can simply not have the `proof` key. 
Actually the claim created on the verification service doesn't really need to contain the `code` param.

c-castillo:
@oed updated

PR #316, opened 2019-02-05
Feat/es6 modules

zachferland:


oed:
@zachferland we should pick this up again. Let's see if we have any spare cycles for it next sprint.

zachferland:
Do you see it supporting or blocking anything? it didn't minimize the size too much. I would wait at least another sprint or two.

oed:
I haven't had time to check actually. You're right, it's not really a rush. Just wanted to make sure that we are not forgetting it 🙂

PR #322, opened 2019-02-06
chore(repo): remove KEY-CONVENTIONS file as it's been added to the 3b…

oed:
…ox main repo


oed:
Uh, this needs to be refactored

PR #328, opened 2019-02-16
Fortmatic integration via ethereum.sendAsync payload id and jsonrpc key

rej156:
BOOM offchain profiles with personal_sign and no metamask!

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -30,7 +30,7 @@ module.exports = {
     var method = 'personal_sign'
     return new Promise((resolve, reject) => {
       ethereum.sendAsync({
-				jsonrpc: "2.0",
+	jsonrpc: "2.0",
         id: new Date().getTime(),
         method,
         params,
@@ -53,7 +53,7 @@ module.exports = {
     var method = 'personal_sign'
     return new Promise((resolve, reject) => {
       ethereum.sendAsync({
-				jsonrpc: "2.0",
+	jsonrpc: "2.0",
         id: new Date().getTime(),
         method,
         params,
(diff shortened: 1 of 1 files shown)


src/utils/index.js
```diff
@@ -28,6 +30,8 @@ module.exports = {
     var method = 'personal_sign'
     return new Promise((resolve, reject) => {
       ethereum.sendAsync({
+	jsonrpc: "2.0",
```
oed: Missing a white space here 🙂

src/utils/index.js
```diff
@@ -49,6 +53,8 @@ module.exports = {
...
```
oed: Same as above

rej156:
https://github.com/ethereum/web3.js/issues/1967

No idea really what the payload id is if I'm honest, seems internal and is reported here
Feels like an implementation detail for concurrency I guess

jsonrpc is very standard yup

Will amend this PR!
I also think we need to change the fromAddress key to from for standardisation too infact

oed:
So does `id` need to be the date, or can it just be `0` or whatever?

Do you know where the definition of `personal_sign` resides online?

rej156:
Looks like a web3.js niggle.
https://github.com/ethereum/go-ethereum/wiki/Management-APIs#personal_sign
In our kauri repo, we set it as Date but yeah it can be 0 or whatever

oed:
Ok, I think I would prefer just `0` then, unless there is a good reason to use Date.

Strange, so in that documentation it says `account`, `fromAddress` is currently being used, and you suggested `from`. I think I'd rather just stay with what we have right now until it becomes a problem 🙂

rej156:
Np makes sense! Will amend based on what you said after this coffee

oed:
Thanks 😊

rej156:
@oed done, sorry not sorry about the prettier plugin on my editor 🥇

oed:
Great, thanks @rej156 !

PR #347, opened 2019-03-17
feat(space): Implemented simple threads

oed:
This is still a WIP but should work quite well already. There is a new `Thread` class. Instances of this class can be created using the Space api. 

Since threads are not added to the users rootstore we need a new way to announce the thread to the pinning node. To do this the new `SYNC_DB` message is used according to our updated pinning node spec. Note that this requires an update to our pinning node in order to function properly

### DIFF
--- a/src/space.js
+++ b/src/space.js
@@ -1,8 +1,10 @@
 const KeyValueStore = require('./keyValueStore')
+const Thread = require('./thread')
 const { sha256Multihash } = require('./utils')
 
 const ENC_BLOCK_SIZE = 24
 const nameToSpaceName = name => `3box.space.${name}.keyvalue`
+const namesTothreadName = (spaceName, threadName) => `3box.thread.${spaceName}.${threadName}`
 
 class Space {
   /**
@@ -12,6 +14,8 @@ class Space {
     this._name = name
     this._3id = threeId
     this._store = new KeyValueStore(orbitdb, nameToSpaceName(this._name), ensureConnected, this._3id)
+    this._orbitdb = orbitdb
+    this._activeThreads = {}
     this._rootStore = rootStore
     /**
      * @property {KeyValueStore} public         access the profile store of the space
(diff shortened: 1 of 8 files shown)


src/space.js
```diff
@@ -40,11 +45,69 @@ class Space {
...
+   *
+   * @param     {String}    name                    The name of the thread
+   * @param     {Object}    opts                    Optional parameters
+   * @param     {Boolean}   opts.noAutoSub          Disable auto subscription to the thread when posting to it (default false)
+   *
+   * @return    {Thread}                            An instance of the thread class for the joined thread
+   */
+  async joinThread (name, opts = {}) {
+    if (this._activeThreads[name]) return this._activeThreads[name]
+    const subscribeFn = opts.noAutoSub ? () => {} : this.subscribeThread.bind(this, name)
```
zachferland: why not subscribe here (or few lines below), whats the thought behind a user must post to a thread to subscribe?

src/thread.js
```diff
@@ -0,0 +1,107 @@
...
+   * @param     {Function}  newPostFn               The function that will get called
+   */
+  async onNewPost (newPostFn) {
+    this._requireLoad()
+    this._db.events.on('replicate.progress', (address, hash, entry, prog, tot) => {
+      let post = entry.payload.value
+      post.postId = hash
+      if (prog === tot) {
+        newPostFn(post)
+        this._queuedNewPosts.map(newPostFn)
```
zachferland: whats the thought behind queuing the posts vs calling newPostFn on each event?

src/thread.js
```diff
@@ -0,0 +1,107 @@
...
+   * @param     {Object}    message                 The message
+   * @return    {String}                            The postId of the new post
+   */
+  async post (message) {
+    this._requireLoad()
+    this._subscribe()
+    this._ensureConnected(this._address, true)
+    return this._db.add({
+      author: this._3id.getDid(),
+      message,
```
zachferland: do we want any max length vals?

src/space.js
oed: You might want to open threads and see the content without needing to subscribe and always look for updates in that thread.

src/thread.js
oed: If you get multiple entries at once from a node, they will be in the wrong order. That's why I'm using `unshift` to add posts to the array.

src/thread.js
oed: That might be a good idea! Do you know of a good way to implement that?

michaelsena: that's interesting, I was thinking about that the other day. Maybe something large - i think Disqus uses 20,000 which is like 6 pages single spaces on MSFT word

oed: We are not enforcing the use of strings though :)

src/space.js
zachferland: oh ok, yeah that makes sense, just the name seems to imply subscribing vs something like openThread

src/thread.js
zachferland: just received in reverse order?

src/thread.js
zachferland: just string length, but actually doesn't matter in this first version, wait to till later and we havent done the same for other vals yet

src/thread.js
oed: yep!

src/thread.js
oed: My point is that we don't enforce it to be a string anywhere, so it might be an object, an array, or anything.

zachferland: right, but doesn't orbit eventually cast to string? or bytes? would cast the same as orbit and do length

zachferland: and I guess maybe we just want strings for threads? not sure

michaelsena: I think we want mixed media threads, so not just "strings" but I imagine as Zach said, Orbit boils everything down to strings in the end, right?

oed: I'm not actually sure about the internals of orbitdb here.

PR #349, opened 2019-03-19
fix: 3id serialize normalized address

zachferland:
Marked do not merge until done following up, with fix, same stores are derived in trust wallet now, but seeing other issues (might be with state there already)

### DIFF
--- a/src/3id/__tests__/3id.test.js
+++ b/src/3id/__tests__/3id.test.js
@@ -1,4 +1,5 @@
 const ThreeId = require('../index')
+const localstorage = require('store')
 
 jest.mock('../../utils/index', () => {
   const sha256 = require('js-sha256').sha256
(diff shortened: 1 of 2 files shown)


zachferland:
Also anyone who created a profile in trust wallet, would lose that store when we make this change

zachferland:
should probably also renormalize whats in local storage already and rewrite as well, otherwise if you tried to open in trustwallet already it would not work

zachferland:
unless recent change from trust wallet, than only profile created since that change would be lost, must be minimal to none

oed:
Oh damn! This completely explains why profiles are lost in trust wallet! We're just lucky that it hasn't happened elsewhere.

zachferland:
yup, can just go into develop 👍 

also added a test for this case

PR #351, opened 2019-03-20
fix: throw during fetch & return empty profile

lsenta:
- [x] _linkProfile => preserve behavior
- [x] _publishRootStore => preserve behavior
- [x] getRootStoreAddress => rethrow
- [x] listSpaces => catch and return an empty array
- [x] getSpace => catch and return an empty object
- [x] getProfiles => rethrow (it's a list of profile so a "global" 404 error should be raised as an error).
- [x] getProfile => empty profile

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -39,14 +39,12 @@ async function getSpace (address, name, serverUrl = PROFILE_SERVER_URL) {
 }
 
 async function getProfile (address, serverUrl = PROFILE_SERVER_URL) {
-  return new Promise(async (resolve, reject) => {
-    try {
-      const res = await utils.fetchJson(serverUrl + '/profile?address=' + encodeURIComponent(address))
-      resolve(res)
-    } catch (err) {
-      reject(err)
-    }
-  })
+  try {
+    // we await explicitly here to make sure the error is catch'd in the correct scope
+    return await utils.fetchJson(serverUrl + '/profile?address=' + encodeURIComponent(address))
+  } catch (err) {
+    return {} // empty profile
+  }
 }
 
 async function getProfiles (addressArray, opts = {}) {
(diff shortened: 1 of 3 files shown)


lsenta:
@oed I modified the fetch function so that it throws when we get a response that has an error status (not 2xx). 

There are a few other functions that are using fetch. I guess we want a consistent behavior for the function that are part of the public API and we want to throw when it's an internal error, so:

- `_linkProfile` => rethrow
- `_publishRootStore` => rethrow
- `getRootStoreAddress` => rethrow
- `listSpaces` => catch and return an empty array
- `getSpace` => catch and return an empty object
- `getProfiles` => rethrow (it's a list of profile so a "global" 404 error should be raised as an error).

src/api.js
```diff
@@ -39,14 +39,12 @@ async function getSpace (address, name, serverUrl = PROFILE_SERVER_URL) {
...
-    } catch (err) {
-      reject(err)
-    }
-  })
+  try {
+    // we await explicitly here to make sure the error is catch'd in the correct scope
+    return await utils.fetchJson(serverUrl + '/profile?address=' + encodeURIComponent(address))
+  } catch (err) {
+    return {} // empty profile
+  }
```
oed: Heh, not sure why there was a promise here before 😅

lsenta:
> Looks good. My main concern is with two of the methods you mentioned.
> `_linkProfile` and `_publishRootStore` Both of these should catch the error and fail silently, as that's the behavior they had before, and we dont wan't these methods failing stoping the user from interacting with their 3box.

I'm on the "fail fast, fail hard" side, if the function is critical we should iron it out. I'll keep the previous behavior (fail silently), could we plan to add some error logging ala sentry.io?

oed:
I'm not familiar with sentry.io :) 
The main reason I'm not so worried about this is that I think we can remove these two requests when we fold the address-server into the pinning-node!

lsenta:
@oed The code is ready for review (I still can't make the test run locally, but looks fine on circleci).

PR #353, opened 2019-03-20
Release/1.5.1

zachferland:


zachferland:
Looks like develop/master history diverged? need to reconcile somehow

zachferland:
has only new commits diffs, but then shows 6 other commits ....

PR #360, opened 2019-03-25
feat(verify): add ethereum addr verif

lsenta:
- [x] implement unit tests
- [x] implement e2e tests
- [x] verify security (could I inject a proof with a different signature and act as someone else?)


oed:
Github says there are conflicts also. Mind rebaseing on develop?

lsenta:
@oed there's no conflict on side, I'm investigating the tests (tricky with my local setup that won't test).

FWIW the issue is there:

```
    > 482 |     expect(Box.getVerifiedAccounts(profile)).rejects.toEqual(new Error())
```

Which is quite confusing me since the method would never fail.
https://github.com/3box/3box-js/blob/develop/src/api.js#L112

PR #363, opened 2019-03-28
Release/1.6.0

oed:
* Feature: Experimental support for threads
* Feature: Add support for DIDs in getProfile and getSpace
* Fix: Handle errors in getProfile correctly
* Fix: Better logic for linking profile to ethereum address


oed:
Uh, looks like we have the same issue with merging this as last release @zachferland

zachferland:
yeah this was expected, didn't permanently fix last time. We are going to have to merge develop into master and master into develop i think. 

But maybe not for the release thought. We could do it after, and then do a squash and merge, so that we don't have all the redundant commits, even thought that squash commit will have redundant changes.

oed:
So:
1. Release this, merge it with master.
2. Merge master to develop
3. Merge develop to master

Is that the right way to permanently fix this? I'm still confused as to what is actually causing this.

zachferland:
I mean you could rebase this on master, to remove extra commits and resolve conflicts. So that would looker cleaner. And then after release merge master in develop (but squash, will still have extra commits, and changes, but at least history will stay readable). And then merge master into develop. 

Pretty sure it it was the master merge into develop before, probably have look at visual since can't quite tell exactly what it did, or why they did converge again later after more merges.

PR #364, opened 2019-03-28
feat(kvstore): add getMetadata on public / private stores (returns timestamp)

lsenta:


src/__tests__/keyValueStore.test.js
```diff
@@ -84,6 +84,25 @@ describe('KeyValueStore', () => {
...
+  describe('metdata', () => {
+    it('should contain the metadata method', async () => {
+      await keyValueStore.set('some-key', 'some-value')
+
+      const v = await keyValueStore.get('some-key')
+      const m = await keyValueStore.getMetadata('some-key')
+
+      expect(v).toEqual('some-value')
+      expect(m).toBeDefined()
+      console.error(m)
```
oed: Why is this console.error here?

src/keyValueStore.js
```diff
@@ -49,6 +59,17 @@ class KeyValueStore {
     return true
   }
 
+  /**
+   * Get the raw value of the given key
+   *
```
oed: Can you add `@private` here so this doen't show up in the api documentation.

src/__tests__/keyValueStore.test.js
lsenta: Forgotten debug, sorry about that!

PR #386, opened 2019-04-11
Add idUtils

lsenta:


src/3box.js
```diff
@@ -11,6 +11,8 @@ const PrivateStore = require('./privateStore')
 const Verified = require('./verified')
 const Space = require('./space')
 const utils = require('./utils/index')
+const { isMuportDID } = require('./utils/id')
+const idUtils = require('./utils/id')
```
oed: Why import twice here?

src/3box.js
```diff
@@ -237,7 +239,7 @@ class Box {
   }
 
   static async _getProfileOrbit (address, opts = {}) {
-    if (utils.isMuportDID(address)) {
+    if (isMuportDID(address)) {
```
oed: why not `idUtils.isMuportDID`?

src/3box.js
```diff
@@ -496,4 +498,6 @@ async function initIPFS (ipfs, iframeStore, ipfsOptions) {
   }
 }
 
+Box.idUtils = idUtils
```
oed: 👍

src/__tests__/idUtils.test.js
```diff
@@ -0,0 +1,26 @@
...
+    expect(isMuportDID('abc')).toEqual(false)
+    expect(isMuportDID('did:example')).toEqual(false)
+    expect(isMuportDID('did:muport')).toEqual(false)
+    expect(isMuportDID('did:muport:Qmb9E8wLqjfAqfKhideoApU5g26Yz2Q2bSp6MSZmc5WrNr')).toEqual(true)
+  })
+
+  test('isClaim', async () => {
+    expect(await isClaim(CLAIM_1)).toEqual(true)
+    expect(await isClaim(INVALID_CLAIM_FORMAT)).toEqual(false)
+    expect(await isClaim(EXPIRED_CLAIM)).toEqual(true) // invalid claim will throw during verify
```
oed: Thanks for testing this!

src/utils/id.js
```diff
@@ -0,0 +1,21 @@
...
+    } catch (e) {
+      return false
+    }
+  },
+  verifyClaim: async (claim, opts = {}) => {
+    try {
+      return await didJWT.verifyJWT(claim, opts)
+    } catch (e) {
+      return null
+    }
```
oed: Let's get rid of the try/catch and just return the result of `verifyJWT` directly!
Will be a pain to debug if invalid claims just return null regardless of the errors.

src/3box.js
lsenta: Fixed!

src/3box.js
lsenta: Fixed!

PR #387, opened 2019-04-11
Feat(network): Send did to pinning node on PIN_DB

oed:
This sends along the DID of the user to the pinning network so that it can be persisted.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -93,7 +93,7 @@ class Box {
 
     const onNewPeer = async (topic, peer) => {
       if (peer === this.pinningNode.split('/').pop()) {
-        this._pubsub.publish(PINNING_ROOM, { type: 'PIN_DB', odbAddress: rootStoreAddress })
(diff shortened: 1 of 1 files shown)


src/3box.js
```diff
@@ -93,7 +93,7 @@ class Box {
 
     const onNewPeer = async (topic, peer) => {
       if (peer === this.pinningNode.split('/').pop()) {
-        this._pubsub.publish(PINNING_ROOM, { type: 'PIN_DB', odbAddress: rootStoreAddress })
+        this._pubsub.publish(PINNING_ROOM, { type: 'PIN_DB', odbAddress: rootStoreAddress, did: this._3id.getDid() })
```
lsenta: We do another `PIN_DB` in the `_ensurePinningNodeConnected` method, should we send the did from there too?

oed: I don't think that is really needed. When that is called we know that the DID will already have been sent to the pinning node once!

PR #397, opened 2019-04-23
fix: use seconds timestamps in metadata

### LINKED ISSUE #396: Timestamp metadata from getSpace() && box.public.get() return 13 digit UNIX timestamp, 3Box Activity library returns 10 digit UNIX timestamp
Not a bug but should probably be streamlined to one!

lsenta:
Fix #396


lsenta:
The approach I took is to fix it in the pinning service so the data is consistent between the lib and the rest API.

We'll have to do a release of the pinning node at the same time: else the response will be inconsistent depending on whether you get metadata through orbit or through the API

oed:
Ok, yeah that makes sense I guess :)

PR #406, opened 2019-04-29
Release/1.7.2

oed:
# Release Notes

## v1.7.2 - 2019-04-29
* Fix: Don't allow setting values wihtout a 'key'
* Fix: Ensure that linkProfile only happens once

Ugh, the commit issues are so annoying. Seems like there is no way around it?


oed:
I have one additional fix that I'd like to get into this release! Let's hold off for now.

oed:
To be rebased on #407

PR #407, opened 2019-04-29
fix(3id): Use orbit-db-keystore to import private key in correct format

oed:
Sometimes we get a different version of elliptic than orbit-db-keystore which causes issues since it thinks that a ec key instance is some other object. This PR fixes this by using orbit-db-keystore directly.

### DIFF
--- a/src/3id/keyring.js
+++ b/src/3id/keyring.js
@@ -1,8 +1,6 @@
 const { HDNode } = require('ethers').utils
 const nacl = require('tweetnacl')
 nacl.util = require('tweetnacl-util')
-const EC = require('elliptic').ec
-const ec = new EC('secp256k1')
 const SimpleSigner = require('did-jwt').SimpleSigner
 const { sha256 } = require('../utils/index')
 
@@ -60,7 +58,7 @@ class Keyring {
   }
 
   getDBKey () {
-    return ec.keyFromPrivate(this.signingKey.privateKey.slice(2))
+    return this.signingKey.privateKey.slice(2)
   }
 
   getDBSalt () {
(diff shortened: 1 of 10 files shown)


src/3id/keyring.js
```diff
@@ -60,7 +58,7 @@ class Keyring {
   }
 
   getDBKey () {
-    return ec.keyFromPrivate(this.signingKey.privateKey.slice(2))
+    return this.signingKey.privateKey.slice(2)
```
zachferland: is this supposed to be the priv key or pub key?

oed: it's supposed to be the priv key since it's passed to orbitdb for writing new entries.

zachferland: ok, yeah didn't read through rest of code not in diff, just wondering what ec.keyFromPrivate was doing before then

oed: The new lines which do `importFromPrivate` basically calls the same function, but with the correct elliptic instance.

zachferland: ok yeah, and you could have put this line here as well to the same effect `await this._orbitdb.keystore.importPrivateKey(dbKey)`

oed: Well yeah, if we pass orbitdb I guess. Or pass the keystore instance?

zachferland: oh I see, yeah I think is fine 👍

oed: that this impl is fine or that we should pass the keystore?

zachferland: this implementation

oed: ok great  😅

oed:
Oh yes! Tests are finally working consistently! Now all ipfs instances are properly closed. Still a bunch of cleanup that could be done on how we handle ipfs in the various tests, but now its functional :D

zachferland:
Nice! yeah that before/after each/all makes a lot more sense now

PR #409, opened 2019-04-29
Feature/set multiple fields

Schwartz10:
Hello! A couple small things to note:

- thoughts on what to return for `setMultiple`? I noticed the tests seem to return `true` if `set` completes without errors? We could do a similar thing for `setMultiple`, but it also might be nice to return the full, updated profile. https://github.com/3box/3box-js/compare/develop...openworklabs:feature/set-multiple-fields?expand=1#diff-efe243c83b3a509aaca8757b8cc08d35R32
- I couldn't get all the tests to run successfully (something on my machine seemed to throw off one of the tests in), but all tests in the files I changed are passing.

### DIFF
--- a/src/__tests__/privateStore.test.js
+++ b/src/__tests__/privateStore.test.js
@@ -88,6 +88,17 @@ describe('PrivateStore', () => {
     expect(privateStore._decryptEntry(retreived)).toEqual('12345')
   })
 
+  it('should set multiple values correctly', async () => {
+    await privateStore._load()
+    await privateStore.setMultiple(['key4', 'key5'], ['yoyo', 'ma'])
+    let dbkey = privateStore._genDbKey('key4')
+    let retreived = privateStore._db.get(dbkey)
+    expect(privateStore._decryptEntry(retreived)).toEqual('yoyo')
+    dbkey = privateStore._genDbKey('key5')
+    retreived = privateStore._db.get(dbkey)
+    expect(privateStore._decryptEntry(retreived)).toEqual('ma')
+  })
+
   it('should get values correctly', async () => {
     let value = await privateStore.get('key1')
     expect(value).toEqual('value1')
@@ -97,6 +108,12 @@ describe('PrivateStore', () => {
 
     value = await privateStore.get('key3')
     expect(value).toEqual('12345')
+
+    value = await privateStore.get('key4')
+    expect(value).toEqual('yoyo')
+
+    value = await privateStore.get('key5')
+    expect(value).toEqual('ma')
   })
 
   it('should remove values correctly', async () => {
--- a/src/keyValueStore.js
+++ b/src/keyValueStore.js
@@ -1,4 +1,4 @@
-const { throwIfUndefined } = require('./utils/index')
+const { throwIfUndefined, throwIfNotEqualLenArrays } = require('./utils/index')
 
 class KeyValueStore {
   /**
@@ -55,6 +55,30 @@ class KeyValueStore {
     const timeStamp = new Date().getTime()
     await this._db.put(key, { value, timeStamp })
     return true
+  }
+
+  /**
+  * Set multiple values for multiple keys
+  *
+  * @param     {Array<String>}    keys                     the keys
+  * @param     {Array<String>}    values                   the values
+  * @return    {Boolean}                                  true if successful, throw error if not
+  */
+  async setMultiple (keys, values) {
+    throwIfNotEqualLenArrays(keys, values)
+    this._requireLoad()
+    this._ensureConnected()
+    try {
+      await keys.reduce(async (previousPromise, nextKey, index) => {
+        await previousPromise
+        throwIfUndefined(nextKey, 'key')
+        const timeStamp = new Date().getTime()
+        return this._db.put(nextKey, { value: values[index], timeStamp })
+      }, Promise.resolve())
+      return true
+    } catch (error) {
+      throw new Error(error)
+    }
   }
 
   /**
--- a/src/privateStore.js
+++ b/src/privateStore.js
@@ -26,6 +26,13 @@ class PrivateStore extends KeyValueStore {
     return super.set(key, value)
   }
 
+  async setMultiple (keys, values) {
+    utils.throwIfNotEqualLenArrays(keys, values)
+    const dbKeys = keys.map(this._genDbKey, this)
+    const encryptedValues = values.map(this._encryptEntry, this)
+    return super.setMultiple(dbKeys, encryptedValues)
+  }
+
   async remove (key) {
     key = this._genDbKey(key)
     return super.remove(key)
--- a/src/publicStore.js
+++ b/src/publicStore.js
@@ -1,5 +1,5 @@
 const KeyValueStore = require('./keyValueStore')
-const { throwIfUndefined } = require('./utils/index')
+const { throwIfUndefined, throwIfNotEqualLenArrays } = require('./utils/index')
 
 class ProfileStore extends KeyValueStore {
   constructor (orbitdb, name, linkProfile, ensureConnected, _3id) {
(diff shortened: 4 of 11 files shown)


src/__tests__/privateStore.test.js
```diff
@@ -86,6 +86,14 @@ describe('PrivateStore', () => {
...
     retreived = privateStore._db.get(dbkey)
     expect(privateStore._decryptEntry(retreived)).toEqual('12345')
+
+    await privateStore.setMultiple(['key4', 'key5'], ['yoyo', 'ma'])
+    dbkey = privateStore._genDbKey('key4')
+    retreived = privateStore._db.get(dbkey)
+    expect(privateStore._decryptEntry(retreived)).toEqual('yoyo')
+    dbkey = privateStore._genDbKey('key5')
+    retreived = privateStore._db.get(dbkey)
+    expect(privateStore._decryptEntry(retreived)).toEqual('ma')
```
oed: Could you put these into a separate test block?

src/publicStore.js
```diff
@@ -12,6 +12,16 @@ class ProfileStore extends KeyValueStore {
     this._linkProfile()
     return super.set(key, value)
   }
+
+  async setMultiple(keys, values) {
```
oed: I think this method should live in the `KeyValueStore` class with a minimal wrapper here.

Schwartz10:
@oed I'm struggling to get the `privateStore` tests to pass at the moment. It's odd because the part it's failing at is [here](https://github.com/3box/3box-js/pull/409/files#diff-5073a726ac80b110d26fdbcb6fd2a684R95). I've verified that it's not a race condition.

Wondering if there is something else I have to adjust in the mocks? Could you provide a second pair of eyes on this one?

With respect to the `setMultiple` for public and private spaces, is [here](https://github.com/3box/3box-js/blob/develop/src/space.js#L121) (public) and [here](https://github.com/3box/3box-js/blob/develop/src/space.js#L172) (private) ok?

Schwartz10:
@oed i fixed the tests. But do let me know about where to write the functionality for `setMultiple` for public and private spaces (see comment above for my best guess)

src/privateStore.js
```diff
@@ -26,6 +26,13 @@ class PrivateStore extends KeyValueStore {
     return super.set(key, value)
   }
 
+  async setMultiple (keys, values) {
+    utils.throwIfNotEqualLenArrays(keys, values)
+    const dbKeys = keys.map(this._genDbKey, this)
+    const encryptedValues = values.map(this._encryptEntry, this)
```
oed: Cool, didn't know you could pass `this` like that. I've always just used `.bind`

src/keyValueStore.js
```diff
@@ -57,6 +57,30 @@ class KeyValueStore {
     return true
   }
 
+  /**
+  * Set multiple values for multiple keys
+  *
+  * @param     {Array<String>}    key                     the keys
+  * @param     {Array<String>}    value                   the values
```
oed: Update the variable names here (add `s`)

Schwartz10:
Hey @oed ! Everything seems to be working well. Tests are passing, and running all these methods for both public/private spaces and public/private profile stores in a real application is working too. How should i proceed?

oed:
Awesome @Schwartz10 I'll have another look at it first thing tomorrow. I assume it should be good, then It'll get merged to develop. Can probably make another release tomorrow as well, if it's urgent 🙂

src/keyValueStore.js
```diff
@@ -57,6 +57,30 @@ class KeyValueStore {
...
+  async setMultiple (keys, values) {
+    throwIfNotEqualLenArrays(keys, values)
+    try {
+      await keys.reduce(async (previousPromise, nextKey, index) => {
+        await previousPromise
+        throwIfUndefined(nextKey, 'key')
+        this._requireLoad()
+        this._ensureConnected()
+        const timeStamp = new Date().getTime()
+        return this._db.put(nextKey, { value: values[index], timeStamp })
```
oed: Why not call `this.set` instead? Am I missing something?

src/space.js
```diff
@@ -122,6 +122,19 @@ const publicStoreReducer = (store) => {
...
     },
+    setMultiple: async (keys, values) => {
+      throwIfNotEqualLenArrays(keys, values)
+      try {
+        await keys.reduce(async (previousPromise, nextKey, index) => {
+          await previousPromise
+          throwIfUndefined(nextKey, 'key')
+          return store.set(PREFIX + nextKey, values[index])
+        }, Promise.resolve())
+        return true
```
oed: Why not do `keys = keys.map(e => PREFIX + e)` then call `store.setMultiple`?

src/space.js
```diff
@@ -170,6 +183,19 @@ const privateStoreReducer = (store, keyring) => {
...
     set: async (key, value) => store.set(dbKey(key), encryptEntry({ key, value })),
+    setMultiple: async (keys, values) => {
+      throwIfNotEqualLenArrays(keys, values)
+      try {
+        await keys.reduce(async (previousPromise, nextKey, index) => {
+          await previousPromise
+          throwIfUndefined(nextKey, 'key')
+          return store.set(dbKey(nextKey), encryptEntry({ key: nextKey, value: values[index] }))
+        }, Promise.resolve())
+        return true
```
oed: Same question as above, but with an additional map? :)

src/keyValueStore.js
Schwartz10: Weird! thought I left a comment about this. 

For some reason running the private store, calling `this.set` did not save the database entries. Not in the tests or in my actual application. As far as i could tell, it didn't have to do with race conditions, the `this` context, or anything else I investigated. I was (and still am) super confused, especially because it worked fine for the public store. [This commit](https://github.com/3box/3box-js/pull/409/commits/77ae568d775c9af23f2815823d9fa66e15c5fe47) solved the problem. Do you have any idea why that would be the case?

src/space.js
Schwartz10: Made this change, but what happens if key is a number? Is it converted to a string somewhere?

src/keyValueStore.js
Schwartz10: I just verified this again in my actual application - returning `this.set` from within `setMultiple` causes private fields to return `null`. SO weird. Would love to know why this is happening.

returning `this._db.put` from inside `this.set` doesn't solve it. 

And old-school promisifying `this.set` doesn't do it either:

```js
set (key, value) {
    return new Promise(async (resolve, reject) => {
      throwIfUndefined(key, 'key')
      this._requireLoad()
      this._ensureConnected()
      const timeStamp = new Date().getTime()
      await this._db.put(key, { value, timeStamp })
      resolve(true)
    })
  }
```

Schwartz10:
Happy to rebase this, but want to wait just before we're ready to pull it in

src/keyValueStore.js
oed: Oh, I think it's because `this` in that context might be the subclass and not the superclass! I.e. it uses privateStore `set` which tries to encrypt the entry again. Then when you get the entry it tries to decrypt it once, when it's encrypted two times.

oed: So yes, we should definitely not call `this.set` then. I think that's better though. We can move `this._requireLoad()` and `this._ensureConnected()` to be called only once in `setMultiple` 😃

src/space.js
oed: 👍 Here it will be converted to a string since we `+` with a string. However it's a good question if a number is passed into orbitdb as a key. I'm not completely sure about that.

Schwartz10:
@oed all set! do you want me to squash this into 1 commit before I rebase?

oed:
Sweet @Schwartz10 and yes that would be nice!

Schwartz10:
@oed all set!

PR #412, opened 2019-05-01
Typescript binding(s) for 3Box

kuhnchris:
Typescript prototypish-ed bindings. Only tested with getProfile right now. Need more usecases for test(s).

### DIFF
--- a/3box.d.ts
+++ b/3box.d.ts
@@ -1,0 +1,107 @@
+declare module "3box"{    
+    interface BoxVerified {
+        Verified: () => BoxVerified,
+        DID: () => string,
+        github: () => any,
+        addGithub: (gistUrl: string) => any,
+        twitter: () => any,
+        addTwitter: (claim: string) => any,
+        email: () => any,
+        addEmail: (claim: string) => any
+    }
+    interface BoxThreadOpts_getPosts {
+        gt?: string,
+        gte?: string,
+        lt?: string,
+        lte?: string,
+        limit?: number,
+        reverse?: boolean
+    }
+    interface BoxThread {
+        BoxThread: () => BoxThread,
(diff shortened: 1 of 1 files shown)


oed:
Hey @kuhnchris! This is great. Would love to merge it, but I'm afraid it will become obsolete quite quickly as the api develops. The reason being that none on our team are very familiar with typescript and it's best practices. 
Maybe it would be better to have this in a separate repo for now, but I'm not sure. What do you think?

kuhnchris:
Hey there @oed - we talked about this on the discord already a bit, but let's recollect here:
The best way would be to commit this to the DefinitlyTyped Repository and tag it with the API version number, that also makes sure that we take into account when the API changes. 

Do you have any policy in place regarding "breaking" changes? 
The most important thing here is to keep at least some kind of accuracy so people can use the bindings. If they need "newer" features they would request it either in the DefinitlyTyped repository, here, or we create a own 3box.ts repository. 
Either way is fine by me, let me know!

Thanks

oed:
Thanks for the summary @kuhnchris. It does seem like committing it to the DefinitlyTyped Repository seems like the best approach indeed. As for breaking changes we follow [Semantic Versioning](https://semver.org/) which should tell you if there are any breaking changes 🙂

oed:
Hey @kuhnchris do you have everything you need here? Can I close this?

kuhnchris:
We can prolly close it, yes. I will get int contact with the DefinitlyTyped repository to get it in. Thanks for the support

oed:
Thanks @kuhnchris!

eshohet:
Any updates on this?

hammadj:
@oed Would love to see updated TS types for 3box

PR #418, opened 2019-05-07
(WIP/Experimental) feat: upgrade orbitdb and ipfs

zachferland:
NOTE: orbitdb identity provider is mocked right now, filling in the rest of of it soon 

These would be changes for the upgrade to orbitdb 0.20.0 and ipfs 0.34.0. To run it includes the following additional changes and the related dependencies. 

-  https://github.com/orbitdb/orbit-db/pull/617
-  https://github.com/orbitdb/orbit-db-access-controllers/pull/31

Allows migration of old stores to new lib. But entries added in new lib will not be able to be parsed in old lib (hardfork). Can see how we will want to handle that.

Still requires more testing, to see if anything else will require changes with upgrades.


zachferland:
Ok found issue, yeah verify identity would not work in orbit right now, signing input (identity.id) is a buffer, while input to verify is not a buffer. Making a PR now, could be in orbit-db-identity-provider or probably better in orbit-db-keystore

zachferland:
https://github.com/orbitdb/orbit-db-keystore/pull/26

oed:
Closing develop is now using orbit-db@0.21

PR #435, opened 2019-05-16
(WIP) Feat/thread access controllers

zachferland:


zachferland:
Still have to work out naming, mostly depends on how we address/name access controllers, and define what makes a unique access controller, thus a unique thread. Moderator access controller name should probably depend on who first mod is, and then would bubble up as different thread addresses

zachferland:
Unique thread access controller is defined by rootMod (first mod), if members, and thread name (which is thread name, plus space). 

This will likely also change how threads are store in space, will want more of this data, so that you have a full reference to thread, or could store the full thread address. Can also change thread name to include this.

src/3id/keyring.js
```diff
@@ -58,7 +64,12 @@ class Keyring {
   }
 
   getDBKey () {
-    return this.signingKey.privateKey.slice(2)
+    return ec.keyFromPrivate(this.signingKey.privateKey.slice(2))
+  }
+
+  async getIdentity () {
```
oed: This would probably make more sense to have in the ThreeId class. That might need some refactoring though. Main reason is that the keyring will not know of the DID of the space.

src/3id/orbitProvider.js
```diff
@@ -0,0 +1,42 @@
...
+    console.log('create new identity')
+    console.log(options)
+    this.pubKey = options.pubKey
+    console.log('pubkey')
+    console.log(this.pubKey)
+  }
+
+  static get type () { return '3ID' } // return type
+  // return identifier of external id (eg. a public key)
+  async getId () {
```
oed: Looks like right now we are creating a separate provider for each identity/space. Is this desired? Another options would be for the `getId` and `signIdentity` function to accept the `options` parameter and include the space name in there. This options can be passed to the `createIdentity` [function](https://github.com/orbitdb/orbit-db-identity-provider/blob/master/src/identities.js#L36).

src/access/thread-open-mod-access.js
```diff
@@ -0,0 +1,135 @@
+'use strict'
+
+const pMapSeries = require('p-map-series')
+const AccessController = require('orbit-db-access-controllers/src/access-controller-interface')
```
oed: This isn't used?

src/access/thread-open-mod-access.js
```diff
@@ -0,0 +1,135 @@
...
+    await this._db.close()
+  }
+
+  async load (address) {
+    if (this._db) { await this._db.close() }
+    // Force '<address>/_access' naming for the database
+    // TODO should memeber and non member threads create different acess controllers
+    // or how to deal with threads with same name with member and non member (different names?
+    //  address of this access controler probably should just change, then different thread name
+    this._db = await this._orbitdb.feed(ensureAddress(address), {
```
oed: What happens here if address is not passed?

src/access/thread-open-mod-access.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /* Private methods */
+  _onUpdate () {
+    // this.emit('updated')
+    // TODO add back
+  }
+
+  /* Factory */
+  static async create (orbitdb, options = {}) {
```
oed: Seems like there are a bunch of things missing here, but I assume you have the entire model figured out?

src/keyValueStore.js
```diff
@@ -141,11 +141,15 @@ class KeyValueStore {
...
-    const key = await this._orbitdb.keystore.importPrivateKey(dbKey)
+    const keyring = this._3id.getKeyringBySpaceName(this._name)
+    const identity = await keyring.getIdentity()
+    const key = keyring.getDBKey()
     this._db = await this._orbitdb.keyvalue(odbAddress || this._name, {
-      key,
-      write: [key.getPublic('hex')]
+      identity,
+      accessController: {
+        write: [key.getPublic('hex')],
```
oed: Shouldn't this be `identity.id`?

src/thread.js
```diff
@@ -75,19 +140,24 @@ class Thread {
...
-    const key = await this._orbitdb.keystore.importPrivateKey(dbKey)
-    this._db = await this._orbitdb.log(odbAddress || this._name, {
-      key,
-      write: ['*']
+    const identity = await this._3id._mainKeyring.getIdentity()
+    this._db = await this._orbitdb.feed(odbAddress || this._name, {
+      identity,
+      accessController: {
+        type: 'thread-access',
+        address: odbAddress,
```
oed: Wait, does this mean that the odbAddress for the accessController is the same as for the thread itself? I thought there would be one feed for the AC and one for the thread itself?
Am I missing something?

zachferland: yeah there is one feed for AC, its created AC, also different since different AC, but cleaned up naming friday (still up for discussion), just pushed up those changes this morning

oed: Oh, now I get what's going on I think 🙂

src/3id/keyring.js
zachferland: yeah makes senes,  this just to get things working, mostly waiting for 3id changes to see exactly how this would look and where, can probably start that now

zachferland:
This first go is probably not sufficient, for our requirements seems like there needs be some causal dependence between the ACL and the data structure. One option we already discussed is to make them part of same struct, otherwise you may be able to have other pointers as well. Many tradeoffs, if separate may to have rebuild entire thread again when receiving AC update (testing each entry against updated AC), if part of the same struct, cant fetch last n entries, need entire struct to build AC. 

Also some links for similar problems/work gathered by ipfs for their crdt work

https://github.com/ipfs/dynamic-data-and-capabilities
https://github.com/protocol/research/issues/8
https://github.com/ipfs-shipyard/peer-star/issues/6
https://github.com/ipfs/dynamic-data-and-capabilities/issues/25

zachferland:
implementation note, access controller right now is feedstore where each entry is {capability, id } 

But options exist here, for more fields, or could be kv store where capability is a key and value is a list of ids, but feedstore seems more efficient.

zachferland:
Now that a thread is also uniquely defined by rootMod and if membersOnly, we could also add that to a a name of thread. Not necessary since that info is also embedded in entire db address by being included in the manifest file. 

But also means we should probably offer the option to open a thread by name or address and should store thread subscription as entire address or with name along with rootMod and if members. Thats all the info needed to reference it.

src/3box.js
```diff
@@ -78,18 +85,23 @@ class Box {
...
     this._ipfs.swarm.connect(this.pinningNode, () => {})
 
+    const keyring = this._3id.getKeyringBySpaceName(rootStoreName)
+    const identity = await keyring.getIdentity()
+    const key = keyring.getDBKey()
     // const cache = (opts.iframeStore && !!cacheProxy) ? cacheProxy : null
-    this._orbitdb = new OrbitDB(this._ipfs, opts.orbitPath) // , { cache })
+    // this._orbitdb = new OrbitDB(this._ipfs, identity, opts.orbitPath) // , { cache })
+    this._orbitdb = new OrbitDB(this._ipfs, identity, {
+      AccessControllers: AccessControllers
```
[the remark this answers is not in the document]
oed: This was an old comment that I missed to submit 🙂

zachferland:
Also need to still inherit from access controller interface to get the event emitters for updates, but some inheritance issues since one goes to es5 and the other is es6 still, need to reconcile those two things

oed:
Is it worth updating the entire 3box codebase to es6 maybe?

zachferland:
@oed  for the comments here that relate to changes for the upgrade, not related to ACs, ill look at those more closely when rebasing after

zachferland:
Most of it is, or do you mean not traspile to es5? I think the problem is that orbit-db-access-controllers is not transpiled, so our transpiled es5 code is trying to inherit from the still es6 class

oed:
Oh, I see. I don't think we need  to inherit from them though. I don't in the AC I wrote.

zachferland:
I don't yet either, and is most everything, the only thing was they also inherit that class from the event emitter, so it has this.emit for emitting update events, for the when the AC gets a new entry, which I suppose revalidates the entires after

oed:
Ok, I think I missed that.

zachferland:
Closed by #455

PR #437, opened 2019-05-16
Feat/create 3id documents

oed:


oed:
Seems like tests are hanging after finishing again :/

zachferland:
@oed were you also creating the orbit identity provider for 3id?

oed:
Yep, working on that today.

oed:
But make that into a separate PR.

src/3id/index.js
```diff
@@ -54,6 +64,54 @@ class ThreeId {
...
+  async _init3ID (spaceName) {
+    const doc = new DidDocument(this._ipfs, DID_METHOD_NAME)
+    if (!spaceName) {
+      const pubkeys = this._mainKeyring.getPublicKeys()
+      doc.addPublicKey('signingKey', 'Secp256k1VerificationKey2018', 'publicKeyHex', pubkeys.signingKey)
+      doc.addPublicKey('encryptionKey', 'Curve25519EncryptionPublicKey', 'publicKeyBase64', pubkeys.asymEncryptionKey)
+      doc.addPublicKey('managementKey', 'Secp256k1VerificationKey2018', 'ethereumAddress', this.managementAddress)
+      doc.addAuthentication('Secp256k1SignatureAuthentication2018', 'signingKey')
+    } else {
+      const pubkeys = this._keyrings[spaceName].getPublicKeys()
```
zachferland: when this is called, there will always be a keyring available by then?

src/thread.js
```diff
@@ -22,7 +22,7 @@ class Thread {
     this._subscribe()
     this._ensureConnected(this._address, true)
     return this._db.add({
-      author: this._3id.getDid(),
+      author: this._3id.muportDID,
```
zachferland: is this just back compat for current threads? but not for threads v2?

src/verified.js
```diff
@@ -6,7 +6,7 @@ class Verified {
    */
   constructor (box) {
     this._box = box
-    this._did = box._3id.getDid()
+    this._did = box._3id.muportDID
```
zachferland: still going to use muport for all verified/claims right now? not just verifying both

src/3id/index.js
```diff
@@ -54,6 +64,54 @@ class ThreeId {
     })
   }
 
+  async _initDID (muportIpfs) {
+    const muportPromise = this._initMuport(muportIpfs)
```
zachferland: Not quite following how both muport and 3id are being used now, so both DIDs are created? and muport is just used for verified, what are the other back compat issues again?

src/thread.js
oed: Yep!

src/verified.js
oed: Yes, it's the simplest thing for now since an ipfs instance is needed to resolve a 3id DID. This would break the `Box.getVerifiedAccounts` method since that doesn't have an ipfs instance.

src/3id/index.js
oed: The rootStore uses the hash of the muport-did for it's name (which was a really bad decision). It also makes sure that the muport is pinned at infura.

oed: It's not going to be super simple to migrate fully, but we'll get there in time!

src/3id/index.js
oed: It should be yes.

Issue #393, opened 2019-04-15
Add passing metadata opts to box.[private/public].get() calls


oznekenzo:
Getting pub/priv data with timestamps requires two calls:
```    
const value = await store.getState().myData.box[type].get(key);
const metadata = await store.getState().myData.box[type].getMetadata(key);

    const valueObject = {
      value,
      timestamp: metadata.timestamp,
    };
```

Would be great if getting this data followed the same pattern and returned data in same format seen elsewhere in lib:
```
      const opts = {
        metadata: true,
      };
      const space = await Box.getSpace(address, spaceName, opts);
```
returns
```
{
 spaceName: {
        key: {
            value: "value",
            timestamp: 123123123123,
}
}
```

oznekenzo:
Same with public.all() & private.all() please!

PR #441, opened 2019-05-19
Optional metadata param for get and all store methods

kimpers:
Implementation of #393

### DIFF
--- a/src/keyValueStore.js
+++ b/src/keyValueStore.js
@@ -15,14 +15,29 @@ class KeyValueStore {
   }
 
   /**
-   * Get the value of the given key
+   * Get the value and optionally metadata of the given key
    *
-   * @param     {String}    key                     the key
-   * @return    {String}                            the value associated with the key, undefined if there's no such key
+   * @param     {String}    key                             the key
+   * @param     {Object}    opts                            optional parameters
+   * @param     {Boolean}   opts.metadata                   return both value and metadata
+   * @return    {String|{value: String, timestamp: Number}} the value associated with the key, undefined if there's no such key
    */
-  async get (key) {
+  async get (key, opts = {}) {
     const x = await this._get(key)
-    return x ? x.value : x
+
+    if (!x) {
+      return x
+    }
+
+    if (opts.metadata) {
+      const metadata = this._extractMetadata(x)
+      return {
+        ...metadata,
+        value: x.value
+      }
+    }
+
+    return x.value
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/keyValueStore.js
```diff
@@ -17,9 +17,22 @@ class KeyValueStore {
...
+
+    if (opts.metadata) {
+      const metadata = this._extractMetadata(x)
+      return {
+        ...metadata,
+        value: x.value
+      }
+    }
+
+    return x.value
```
kimpers: This is a bit funky but I assumed that you'd want keep compatibility with the old api returning a strings when the `metadata` option is not passed, is that a correct assumption?

oed: Hm, I can't remember exactly why we had `x ? x.value : x` rather than just returning null if x is not defined. I think this is clear enough though!

src/keyValueStore.js
```diff
@@ -17,9 +17,22 @@ class KeyValueStore {
    * @param     {String}    key                     the key
```
oed: Could you add the optional param `opts.metadata` to the jsdoc here?

kimpers: Yep definitely, still wip. Just wanted to get some comments on whether I should maintain compatibility with the old format for non metadata queries or if they always should use the new response format.

Having the return be string|object could be confusing for people

oed: Yep, I definitely see the concern. However I think it's not a good idea to break backwards compatibility. 
We have the same quirk in the `Box.getProfile` function as well.

kimpers:
Ready for review now 😄 I tried to keep the changes to a minimum since I'm not familiar with the codebase

src/keyValueStore.js
kimpers: > Cool, Thanks for picking this up @kimpers!
> 
> Looks good, but I think this could have some side effects that needs to be addressed. Specifically in `publicStore` and `privateStore` classes, as well as the public/private reducers in the `space` class.

I think everything should be addressed now

kimpers:
> Cool, Thanks for picking this up @kimpers!
> 
> Looks good, but I think this could have some side effects that needs to be addressed. Specifically in `publicStore` and `privateStore` classes, as well as the public/private reducers in the `space` class.

I think everything should be addressed now 🎉

oed:
Nice, thanks! I'll try to have a look at this by the end of the week or early next week 👍

oed:
Looks great @kimpers! If you can rebase this on develop I can merge it 🙂

kimpers:
> Looks great @kimpers! If you can rebase this on develop I can merge it 🙂

Awesome 🎉  it has been updated now 😃

oed:
Cool, thank you @kimpers!

Issue #439, opened 2019-05-17
call to setMultiple for the first time does not await signature request before resolving


Schwartz10:
**Describe the bug**
On the first time I call `setMultiple` - I need to sign a message. However, the promise to `setMultiple` resolves before the signature requests completes. I'm not sure if this is something to do with the actual method, or the async pattern used in [this method](https://github.com/3box/3box-js/blob/develop/src/keyValueStore.js#L67)

Any ideas why this might be happening? 

**To Reproduce**
1. Start on a new metamask account
2. open your 3box
3. prepare to set a field using the `setMultiple` method (you could just wrap one field in an array)
4. throw debugger keyword after the call to `setMultiple`
5. invoke the call to `setMultiple`

Notice - the signature request is made, but the value `true` is returned before I sign the message.

<img width="1169" alt="image" src="https://user-images.githubusercontent.com/12353734/57950583-b1dd6100-78b5-11e9-81a0-e63509a4c690.png">

You'll notice on the very far right of this screenshot - the aragon signature panel is starting to peak out, yet the promise to `setMultiple` has been resolved.

**Expected behavior**
`setMultiple` should wait until the create box signature is completed before resolving its promise

**Screenshots**
(see above)

**Desktop (please complete the following information):**
 - Browser: chrome
- Version 74.0.3729.157 (Official Build) (64-bit)

oed:
Thanks for the exemplar bug report @Schwartz10 👌 

If I understand correctly this is referring to the "linking signature", i.e. not the signatures that are triggered by `openBox` or `openSpace`.

I think this is because we are not awaiting [this promise.](https://github.com/3box/3box-js/blob/develop/src/publicStore.js#L19) The same issue should also appear when calling `set`. Although this issue should **only** occur on the publicStore.

This was a deliberate design decision, but I totally get that it might not be very desirable. It should be simple to fix though. Would you just prefer the promise to resolve after the signature has been made?

Schwartz10:
> If I understand correctly this is referring to the "linking signature"

Yeah, this occurs after the user's box is already open.

> I think this is because we are not awaiting this promise

That looks like it!

> This was a deliberate design decision, but I totally get that it might not be very desirable. It should be simple to fix though. Would you just prefer the promise to resolve after the signature has been made?

Understood. Yeah - because if the user denies the signature request, the UI makes it seem like their profile was still "created".

Schwartz10:
Ah yes. Manually requesting this linking could be even better

[...]

PR #445, opened 2019-05-21
Hotfix/v1.8.2

zachferland:


oed:
This now also fixes #439

zachferland:
thanks @oed for adding to it, did at the very end of day and didn't double check tests etc. 

account type makes sense 👍  why is isAccountLinked needed? just for clarity?

oed:
Yep, for clarity and if someone wants to check before they call linkAccount.

PR #450, opened 2019-05-23
feat(3id): Create and use 3ID orbitdb-identity-provider

oed:
This PR creates the 3ID provider for orbitdb and uses it for the rootstore as well as the keyvalue stores.
Right now I have disabled the tests for the threads since I have not migrated them to use the new identity provider. 

js-ipfs is updated to 0.35

orbitdb is updated to 0.21-rc
added orbit-db-access-controllers as a dep for now until they are released and will bubble up to orbitdb.

More tests are using snapshots now for simplicity.


zachferland:
Do have some issues with the tests, follow up soon

zachferland:
Oh actually all good, did need this though

```
Error: Cannot find module 'ultron' from 'WebSocket.js'
```

PR #460, opened 2019-05-27
feat(api): add blocklist to getProfile and getSpace

oed:
This PR adds a simple blocklist function that can be passed to `getProfile` and `getSpace`. This function accepts an address and should return a boolean. If true the `get` functions will throw an error saying the user is blocked.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -179,6 +179,8 @@ class Box {
    *
    * @param     {String}    address                 An ethereum address
    * @param     {Object}    opts                    Optional parameters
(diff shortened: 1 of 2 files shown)


zachferland:
don't really get the need for this (?), when dev can call their own function before calling getProfile, but if it conveys something useful and/or minimize the need to explain something then sure 👍

oed:
Yep the intention is just to make it easier and clearer for developers to understand how to use a blocklist. In the future we might also provide standard blocklists.

PR #483, opened 2019-06-06
Feat/by address example up

zachferland:


zachferland:
oh actually gonna add a test or two to this

PR #484, opened 2019-06-06
feat: add onNewCapabilities fn for AC updates

zachferland:
user friendly api for AC member/moderator updates


oed:
Maybe not now, but we should perhaps start using eventemitters for these things

zachferland:
yeah i agree, you mean just exposing that event interface and documenting the events?

oed:
yeah, then we also get unsubscribe for free :)

PR #485, opened 2019-06-07
feat(space): add subDID to root-store entry

oed:
Puts the subDIDs in the root-store entry that also has the space odbAddress. Also adding a `type` to the root-store entry.

### DIFF
--- a/src/space.js
+++ b/src/space.js
@@ -29,6 +29,13 @@ class Space {
     this.private = null
   }
 
+  /**
+   * @property {String} DID        the did of the user in this space
+   */
+  get DID () {
+    return this._3id.getSubDID(this._name)
+  }
+
   async open (opts = {}) {
     if (!this._store._db) {
       // store is not loaded opened yet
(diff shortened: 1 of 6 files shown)


src/space.js
```diff
@@ -122,15 +132,6 @@ class Space {
...
   }
-
-  async _ensureDIDPublished () {
-    // Ensure we self-published our did
-    if (!(await this.public.get('proof_did'))) {
-      await this._syncSpacePromise
-      // we can just sign an empty JWT as a proof that we own this DID
-      await this.public.set('proof_did', await this._3id.signJWT({}, { space: this._name }), { noLink: true })
-    }
-  }
```
zachferland: 👍

src/space.js
```diff
@@ -37,8 +44,12 @@ class Space {
       const spaceAddress = await this._store._load()
 
       const entries = await this._rootStore.iterator({ limit: -1 }).collect()
-      if (!entries.find(entry => entry.payload.value.odbAddress.indexOf(nameToSpaceName(this._name)) !== -1)) {
-        this._rootStore.add({ odbAddress: spaceAddress })
+      const entry = entries.find(entry => entry.payload.value.odbAddress.indexOf(nameToSpaceName(this._name)) !== -1)
+      if (!entry) {
+        await this._rootStore.add({ type: 'space', DID: this.DID, odbAddress: spaceAddress })
```
zachferland: 👍

zachferland: then in future, rootstore <-> space/did link can be verified by resolving the did now? right?

oed: Precisely :)

PR #489, opened 2019-06-10
ref: api changes, joinThreadByAddress, keep active threads by address

zachferland:
rodMod -> firstModerator
membersOnly -> members 
space._activeThreads by address instead of name 
refactor 3id mock for tests, old keyvalue store did mock failed, 3box test still failing, need to update


zachferland:
@oed have tests passing, also add check for valid 3id for capabilities, although those func may also all addresses in future as well

oed:
ok, I think it's fine for now. Unless it's an easy add!

PR #494, opened 2019-06-13
feat: join firstmod address, addMember(Moderator) by address

zachferland:


zachferland:
@oed saw your discord message, but was just to quick to do while all code was still in mind :)

zachferland:
sneaky throw, was just throwing and loggining in the console where ever it felt like, in no relation to the actual test, all set now

PR #495, opened 2019-06-13
ref: await linkprofile before adding data, throw if not signed

zachferland:


src/3box.js
```diff
@@ -488,8 +488,12 @@ class Box {
     if (!linkData) {
       const address = this._3id.managementAddress
       const did = this._3id.muportDID
-
-      const consent = await utils.getLinkConsent(address, did, this._web3provider)
+      let consent
+      try {
+        consent = await utils.getLinkConsent(address, did, this._web3provider)
+      } catch (e) {
+        throw new Error('Link consent message must be signed before adding data, to link address to store')
```
oed: Nice addition!

src/3box.js
```diff
@@ -506,11 +510,7 @@ class Box {
     }
 
     // Send consentSignature to 3box-address-server to link profile with ethereum address
-    try {
-      await utils.fetchJson(this._serverUrl + '/link', linkData)
-    } catch (err) {
-      console.error(err)
-    }
+    utils.fetchJson(this._serverUrl + '/link', linkData).catch(console.log)
```
oed: Why console.log and not console.error? Or maybe warn?

zachferland: yeah just changed that

oed: haha, you fixed it before I finished the review :P

zachferland: we are too quick now!

tests still not happy about it yet

PR #496, opened 2019-06-13
feat: link account -> address for clarity

zachferland:


oed:
Did you have any thoughts on moving them to `box.utils.linkAddress()` or something like that, as mentioned in the main ticket?

PR #500, opened 2019-06-17
feat: store address links in root and v1 format

zachferland:


src/3box.js
```diff
@@ -493,24 +497,29 @@ class Box {
...
-        linked_did: did
+        version: 1,
+        type: ACCOUNT_TYPES.ethereumEOA,
+        message: consent.msg,
+        signature: consent.sig,
+        timestamp: consent.timestamp
       }
-      await this.public.set('ethereum_proof', linkData, { noLink: true })
+
+      const docHash = (await this._ipfs.add(Buffer.from(JSON.stringify(linkData))))[0].hash
```
oed: I'd prefer if we use the DAG API instead of the files API: https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/DAG.md

src/3box.js
```diff
@@ -523,6 +532,32 @@ class Box {
...
+    }
+    await this._rootStore.add(link)
+  }
+
+  async _readAddressLinks() {
+    const entries = await this._rootStore.iterator({ limit: -1 }).collect()
+    const linkEntries = entries.filter(e => e.payload.value.type === 'address-link')
+    return linkEntries.map(async (entry) => {
+      // TODO handle missing ipfs obj??, timeouts?
+      const obj = JSON.parse(await this._ipfs.cat(entry.payload.value.data))
```
oed: Same thing here: https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/DAG.md

src/3box.js
```diff
@@ -523,6 +532,32 @@ class Box {
...
+      data
+    }
+    await this._rootStore.add(link)
+  }
+
+  async _readAddressLinks() {
+    const entries = await this._rootStore.iterator({ limit: -1 }).collect()
+    const linkEntries = entries.filter(e => e.payload.value.type === 'address-link')
+    return linkEntries.map(async (entry) => {
+      // TODO handle missing ipfs obj??, timeouts?
```
oed: We should try to make sure that this doesn't happen. Seems quite difficult to deal with.

src/3box.js
```diff
@@ -523,6 +532,32 @@ class Box {
...
+      if (!obj.address) {
+        obj.address = utils.recoverPersonalSign(obj.message, obj.signature)
+      }
+      return obj
+    })
+  }
+
+  async _readAddressLink(address) {
+    const links = await this._readAddressLinks()
+    return links.filter(link => link.address === 'address-link')[0]
```
oed: `find` instead of `filter` 🙂
Oh, and `'address-link'` should be the `address` param right?

package.json
```diff
@@ -48,6 +48,7 @@
     "3id-resolver": "^0.0.5",
     "@babel/runtime": "^7.4.5",
     "did-jwt": "^0.1.3",
+    "eth-sig-util": "^2.2.0",
```
oed: Not sure if we need to optimize for this now, but this library seems quite heavy for just verifying a signature.

src/3box.js
```diff
@@ -466,6 +467,7 @@ class Box {
    * @param     {String}        type        The type of link (default 'ethereum')
    */
   async linkAddress (type = ACCOUNT_TYPES.ethereum) {
+    // TODO rebase with addresslink, and will this take proof/args or just default to managementAddress? or another func
```
oed: I think it should take proof as an optional param.

src/3box.js
```diff
@@ -25,7 +25,8 @@ const config = require('./config.js')
 const API = require('./api')
 
 const ACCOUNT_TYPES = {
-  ethereum: 'ethereum'
+  ethereum: 'ethereum',
```
oed: Do we need this type still?

src/3box.js
zachferland: yeah i think, do mean assuming a proof is a CID?

src/3box.js
zachferland: 👍 yeah was going to change that

src/3box.js
zachferland: technically if anyone passed it in isAccountLinked, but don't thing it would have been used at all there, could be removed

zachferland:
Main problem for this pr, has been getting the tests to work, tests are not serving their purpose too well when they take many times longer than writing the feature :( and still can't get existing tests to work (did not push any test changes up yet)

src/3box.js
oed: I assume the proof is an object that will be saved into ipfs by this function :)

oed:
Is there maybe some refactoring that we need to do in order to make the 3box.js test less monolithic?

zachferland:
yeah its continually sapping productivity lately, a few things
-  make all the unit test independent, too many rely on conditions setup in prior tests, which is easy to forgot or not obvious when making changes later, all should have conditions laid out in its own test
- less mocks of our code, mostly mock remote calls, external libs, and memory, a lot of overhead when you have to change your own mocks a lot, or just have refactor and clean up mocks better
- still other random errors that make no sense, and thrown un related to any code near them, hard to get good traces within the tests, not sure the fix here

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "./node_modules/.bin/standard --verbose src/**",
-    "test": "rm -rf ./tmp ; jest --forceExit --coverage --runInBand --testURL=\"http://localhost\"",
+    "test": "rm -rf ./tmp && jest --forceExit --coverage --runInBand --testURL=\"http://localhost\"",
```
oed: I thought `;` and `&&` are equivalent?

oed:
I think this mostly looks good otherwise. Still wondering why we send the entry along to the address-server though?

src/3box.js
```diff
@@ -519,10 +536,43 @@ class Box {
...
     utils.fetchJson(this._serverUrl + '/link', linkData).catch(console.error)
   }
 
+  async _writeAddressLink(proof) {
+    const data = (await this._ipfs.dag.put(proof)).toBaseEncodedString()
+    const link = {
+      type: 'address-link',
+      data
+    }
+    const entryCID = await this._rootStore.add(link)
```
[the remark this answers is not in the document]
zachferland: yeah this changed since, but added a check before writes

package.json
zachferland: yeah basically, && just only runs if first command doesn't fail, just wanted to make things actually deleted, thought i had some test issues due to that

oed: I see, great!

src/utils/index.js
```diff
@@ -1,20 +1,32 @@
...
 module.exports = {
   getMessageConsent,
 
+  recoverPersonalSign: (msg, personalSig) => {
+    if (!msg || !personalSig) throw new Error('recoverPersonalSign: missing arguments, msg and/or personalSig')
+    const msgParams = {
+      data: msg,
+      sig: personalSig
+    }
+    return sigUtil.recoverPersonalSignature(msgParams)
```
oed: Would it be a lot of additional work to recover the signature with ethersjs? 
https://docs.ethers.io/ethers.js/html/api-utils.html?highlight=signature#elliptic-curve

oed: I guess we just need to prepend the right data for an eth-signature?

oed: Since we're already importing ethers anyway it would remove the dependency on eth-sig-util

zachferland: yeah good idea, switched over

PR #502, opened 2019-06-18
chore(docs): clarify join thread without opts

oed:


zachferland:
Just modify the template instead, or I can also update tmrrw as well

oed:
Ugh, sorry. Fixed now!

PR #506, opened 2019-06-22
Feat/remove address link

zachferland:
built on other pr

not tested yet with address server, and probably could add one unit test here


src/3box.js
```diff
@@ -461,56 +462,101 @@ class Box {
...
-   * Checks if there is a proof that links an external account to the 3Box account of the user.
+   * Remove given address link, returns true if successful
    *
-   * @param     {String}        type        The type of link (default ethereum)
+   * @param     {String}   address      address that is linked
    */
-  async isAddressLinked (type = ACCOUNT_TYPES.ethereum) {
-    if (type === ACCOUNT_TYPES.ethereum) {
-      return Boolean(await this.public.get('ethereum_proof'))
+  async removeAddressLink (address) {
```
oed: We also need to remove the `'link-address'` entry from the rootStore in this method.

PR #509, opened 2019-07-04
Feat/ref remove address link

zachferland:


zachferland:
Replacing https://github.com/3box/3box-js/pull/506 - replaced, easier than rebasing

PR #512, opened 2019-07-05
Fix/some bugs

oed:
Two bugfixes:

* Added support for web3 1.0 was implemented totally incorrectly and not tested properly. 
* Adding `address-link` entries to the root-store caused openSpace to break

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build": "npm run build:es5; npm run build:dist; npm run build:dist:prod; npm run build:dist:api",
     "prepublishOnly": "npm run build; npm run generate-readme",
     "example-server:start": "node example/server.js",
-    "example:start": "npm run build:dist -- --watch; npm run example-server:start",
+    "example:start": "npm run build:dist -- --watch & npm run example-server:start",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "build": "npm run build:es5; npm run build:dist; npm run build:dist:prod; npm run build:dist:api",
     "prepublishOnly": "npm run build; npm run generate-readme",
     "example-server:start": "node example/server.js",
-    "example:start": "npm run build:dist; npm run example-server:start",
+    "example:start": "npm run build:dist -- --watch & npm run example-server:start",
```
zachferland: ah yeah forgot to change in past, can also use build:dist:dev  for watching / live rebuild

src/space.js
```diff
@@ -44,7 +44,7 @@ class Space {
       const spaceAddress = await this._store._load()
 
       const entries = await this._rootStore.iterator({ limit: -1 }).collect()
-      const entry = entries.find(entry => entry.payload.value.odbAddress.indexOf(nameToSpaceName(this._name)) !== -1)
+      const entry = entries.find(entry => entry.payload.value.odbAddress && entry.payload.value.odbAddress.indexOf(nameToSpaceName(this._name)) !== -1)
```
zachferland: 👍

PR #513, opened 2019-07-05
Feat/list address links

oed:
To be merged after #512


oed:
Thanks, will rebase and merge.

Issue #528, opened 2019-07-18
Add api/option for wallets to provide signature 


danjm:
Hi there,

I'm a member of the Metamask team and currently working on integrating 3box to MetaMask. Our initial release of this will just be used to back up a users settings data.

To provide an optimal onboarding experience, we would like to initiate 3box syncing without the user having to confirm a signature request.

This could be possible if 3box allowed for a wallet signed message to be passed as an option to `openBox`. Currently, threebox gets the signature it needs for authentication through a `personal_sign` request in the `openBoxConsent` call made here: https://github.com/3box/3box-js/blob/develop/src/3id/index.js#L184

Naively, it seems that 3box could do something like this:
```
const sig = opts.walletProvidedSignature
        ? opts.walletProvidedSignature
        : await utils.openBoxConsent(normalizedAddress, ethereum)
```
Provided that the `walletProvidedSignature` meets 3boxes specifications.

This would allow metamask and other signers to use 3box internally without our users having to manually confirm a signature request.

If this is doable, I'm happy to help with associated dev work within 3box-js, and happy to discuss more here or via a call

oed:
Hey @danjm this is a great suggestion! We are actually working on exactly this issue right now. We are building an Identity Wallet SDK that allows to pass either and `authSecret` or a `seed`. You can read more about it here: https://github.com/3box/3box/issues/640

In the meantime, is it possible to automatically approve the request if you see that it originates from within MetaMask?

Happy to discuss more and jump on a call :)

oed:
I guess adding an extra option in `openBox` would be quite simple as a quick solution to this particular pain point for now though. Is that something you'd be interested in contributing?

danjm:
> Is that something you'd be interested in contributing?

Yes, I'm happy to contribute that.

oed:
Cool, I'll have a quick look at this tomorrow morning to add a couple of notes for you to start with!

oed:
Ok, three steps to make this happen I believe:

1. Add optional parameter `conentSignature` (in opts object) to `openBox` in `3box.js` file
1. Add optional parameter `conentSignature` (in opts object) to `getIdFromEthAddress` in `3id/index.js` file
1. In the else block in `getIdFromEthAddress` use `conentSignature` for `sig` variable if defined

I think that's it! 
Let me know if anything is unclear :)

PR #535, opened 2019-07-23
Adds contentSignature option to getIdFromEthAddress

danjm:
Resolves https://github.com/3box/3box-js/issues/528

This PR adds a contentSignature option to the `getIdFromEthAddress` method in `3id.js`

This can enable wallets to provide signatures for authorizing 3box as an option when creating a new 3box instance, thereby removing the need for user action to confirm the signature.

The PR also updates documentation for `openBox` in `3box.js` and adds unit tests to cover the new functionality of `getIdFromEthAddress`

### DIFF
--- a/src/3id/__tests__/3id.test.js
+++ b/src/3id/__tests__/3id.test.js
@@ -25,9 +25,11 @@ const clearLocalStorage3id = (address) => {
 
 const ADDR_1 = '0x12345'
 const ADDR_2 = '0xabcde'
+const ADDR_3 = '0xlmnop'
 const ADDR_1_STATE_1 = '{"managementAddress":"0x12345","seed":"0xbc95bb0aeb7e5c7a9519ef066d4b60a944373ba1163b0c962a043bebec1579ef33e0ef4f63c0888d7a8ec95df34ada58fb739b2a4d3b44362747e6b193db9af2","spaceSeeds":{}}'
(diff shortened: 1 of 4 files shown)


src/3id/__tests__/3id.test.js
```diff
@@ -87,6 +89,36 @@ describe('3id', () => {
...
+    })
+
+    it('should NOT create the same identity given the same address but a different contentSignature', async () => {
+      // did is mocked, so compares serialized state
+      const opts = { contentSignature: '0xsomeIncorrectSignature' }
+      const threeId1 = await ThreeId.getIdFromEthAddress(ADDR_3, ETHEREUM, ipfs, opts)
+      clearLocalStorage3id(ADDR_3)
+      const threeId2 = await ThreeId.getIdFromEthAddress(ADDR_3, ETHEREUM, ipfs, opts)
+      expect(mockedUtils.openBoxConsent).toHaveBeenCalledTimes(0)
+      expect(threeId1.serializeState()).not.toEqual(threeId2.serializeState())
```
oed: I'm not really following why this line would be not equal. Isn't the same consentSignature being used for both ThreeId instances? I also don't see any difference in this test and the previous one. What am I missing?

danjm: The contentSignature on lines 105 and 115 are different:
105: `const opts = { contentSignature: '0xsomeContentSignature' }`
vs.
115: `const opts = { contentSignature: '0xsomeIncorrectSignature' }`

This could be made more clear by using capitalized constants as is done for addresses. I will make that change now.

danjm: Just pushed c7d8815 which should make this clearer

oed: I see that they are different in the two different tests, but in this test uses the opts form line 117 for both `threeId1` and `threeId2` so I don't understand why their serialized states would be different?

oed: @danjm Did you see my last comment? I'm still having trouble understanding this. But I'm probably missing something trivial. Will try to run locally today to get a better understanding.

oed: Oh I see now! It's because it loads the state from localstorage for `threeId1` and then it's cleared for the second one!

danjm: Ah, nice catch. I didn't notice that before... and I guess I didn't really understand your initial comment here until now. Looking at this now, I would write the test differently so it wasn't dependent on that local storage clear. I guess this works for now though. Thanks for following up on that.

oed: I see. Should be fine for now yes. Just started preparing a release for this so you can get on with your integration: https://github.com/3box/3box-js/pull/540 Should be ready by the end of the day I think.

Issue #534, opened 2019-07-23
onSyncDone race conditions


Schwartz10:
**Describe the bug**
If `onSyncDone` is called too late after `Box.openBox`, the call hangs and the registered callback listener never fires

**To Reproduce**
```js
const openedBox = await Box.openBox(ethereumAddress, web3)

setTimeout(() => {
  openedBox.onSyncDone(() => {
    // unreachable code, callback never fires
  })
}, 5000)
```

**Expected behavior**
The callback should fire if the box has already synced

Schwartz10:
An update from @chadoh (another dev who was experiencing hanging calls) - he's on Brave browser and saw these warnings in the console:

Warning: Given input entry is not allowed in this log and was discarded (no write access).

Store.js:258 "WARNING! Head hash didn't match the contents

Warning: Given input entry is not allowed in this log and was discarded (no write access).

[Violation] 'setInterval' handler took 76ms

He can't consistently unlock a box on brave, using code directly from the 3box docs. Either calls hang, or the box unlocks extremely quickly.

Schwartz10:
Just jotting some notes down here, it is possible that the syncing flow gets stuck on [this line](https://github.com/3box/3box-js/blob/develop/src/3box.js#L159). I'm happy to debug further but going to hang tight for now and wait to see if y'all have any ideas

oed:
Seems like what @chadoh is experiencing might be a bit of a different issue. Or what do you mean by "unlock a box" exactly? Sounds like it has something to do with `openBox` not working. Are you referring to `onSyncDone` specifically though?

PR #538, opened 2019-07-23
naive fix for #534

Schwartz10:
Closed a different, identical PR bc of a weird commit history. 

@oed I'm sure this is a pretty naive approach to the race condition issue, but it seems to do the trick for me. I'll check with another dev on the team who was also experiencing hanging calls to see if it fixes it for him too and update here.

Would love to get a fix for this issue in to a version <1.10 because I want to avoid this issue

Please let me know your thoughts! Thank you

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -65,6 +65,7 @@ class Box {
     this._ipfs = ipfs
     this._serverUrl = opts.addressServer || ADDRESS_SERVER_URL
     this._onSyncDoneCB = () => {}
+    this._boxSynced = false
     /**
      * @property {KeyValueStore} public         access the profile store of the users 3Box
      */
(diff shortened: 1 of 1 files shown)


Schwartz10:
An update from @chadoh (another dev who was experiencing hanging calls) - he's on Brave browser and saw these warnings in the console:

```
Warning: Given input entry is not allowed in this log and was discarded (no write access).
```
```
Store.js:258 "WARNING! Head hash didn't match the contents
```
```
Warning: Given input entry is not allowed in this log and was discarded (no write access).
```
```
[Violation] 'setInterval' handler took 76ms
```

He can't consistently unlock a box on brave, using code directly from the 3box docs. Either calls hang, or the box unlocks extremely quickly.

Schwartz10:
Hmmm, also not sure this actually fixes the issue i was seeing before. I still get stuck syncing at some point, but can't seem to nail down why or where.

Will drop some more info in the issue ticket

oed:
This makes sense @Schwartz10. I don't see any issues with it 🙂

Schwartz10:
@oed that's great! When do you think we could pull this in and publish a new version?

PR #545, opened 2019-08-01
fix: refine rootstore sync logic

oed:
Basically we should wait for the rootstore to be synced before we do the `onSyncDone` callback.

This PR has the following changes:
* Listen for `HAS_ENTRIES` message for rootstore and wait until the rootstore is fully synced
* If `numEntries === 0` we know that it's a new user and we call `onSyncDone` directly
* `_createRootStore` which adds the entries for the public and private store is now added after `onSyncDone` has happened (the rootstore is fully synced)

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -145,6 +145,29 @@ class Box {
         setTimeout(() => { spaceMessageFilterActive = false }, 3000)
       }
       if (data.type === 'HAS_ENTRIES') {
+        console.log(data)
+        if (data.odbAddress === rootStoreAddress) {
+          if (data.numEntries === 0) {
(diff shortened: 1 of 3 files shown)


src/3box.js
```diff
@@ -145,6 +145,29 @@ class Box {
         setTimeout(() => { spaceMessageFilterActive = false }, 3000)
       }
       if (data.type === 'HAS_ENTRIES') {
+        console.log(data)
```
oed: This line should be removed before this PR is merged.

zachferland: other logs below as well

oed:
Closing in favor of https://github.com/3box/3box-js/pull/547

PR #554, opened 2019-08-12
feat: openBox now works with IdentityWallet instance

oed:
This PR adds support for passing an IdentityWallet instance to `openBox`. 

This functionality still needs some integration tests to make sure that all IdentityWallet functionality works as intended.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -98,6 +98,12 @@ class Box {
     this.replicator.rootstore.setIdentity(await this._3id.getOdbId())
     this.syncDone = this.replicator.syncDone
 
+    if (this._3id.idWallet) {
+      this._3id.idWallet.events.on('new-auth-method', authData => {
+        this._writeRootstoreEntry(Replicator.entryTypes.AUTH_DATA, authData)
+      })
+    }
+
     this.public = new PublicStore(this._3id.muportFingerprint + '.public', this._linkProfile.bind(this), this.replicator, this._3id)
     this.private = new PrivateStore(this._3id.muportFingerprint + '.private', this.replicator, this._3id)
     await Promise.all([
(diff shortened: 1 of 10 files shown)


src/3box.js
```diff
@@ -96,12 +96,16 @@ class Box {
...
+    }
+
     this.public = new PublicStore(this._3id.muportFingerprint + '.public', this._linkProfile.bind(this), this.replicator, this._3id)
     this.private = new PrivateStore(this._3id.muportFingerprint + '.private', this.replicator, this._3id)
-    await Promise.all([
-      this.public._load(),
-      this.private._load()
-    ])
+    await this.public._load()
+    await this.private._load()
```
zachferland: prior implementation would have been faster in most env, unless you need public load to resolve before private load

src/__tests__/idWallet.integration.js
```diff
@@ -0,0 +1,139 @@
...
+
+  it('should get same state on openBox with IdentityWallet opened using second authSecret', async () => {
+    publishHasEntries()
+    idWallet = new IdentityWallet({ authSecret: AUTH_2, ethereumAddress: ADDRESS })
+    const box = await Box.openBox(idWallet, mockEthProvider, opts)
+    await box.syncDone
+    expect(await box.public.all()).toEqual(pubState)
+    await box.close()
+  })
+})
```
zachferland: 👍

src/__tests__/testUtils.js
```diff
@@ -1,178 +1,37 @@
...
+      Addresses: {
+        Swarm: [
+          `/ip4/127.0.0.1/tcp/${4004 + id * 2}`,
+          `/ip4/127.0.0.1/tcp/${4005 + id * 2}/ws`
+        ],
+        API: `/ip4/127.0.0.1/tcp/${5003 + id}`,
+        Gateway: `/ip4/127.0.0.1/tcp/${9091 + id}`
+      },
+      Bootstrap: []
+    }
```
zachferland: yes! thanks

src/3box.js
oed: I was seeing some weird concurrency issue here when running the integration test. I don't think this would be much slower at all tbh!

oed:
Needs a rebase, then should be ready for merge.

PR #570, opened 2019-08-16
Hotfix/v1.10.7

oed:
# Release Notes

## v1.10.7 - 2019-08-17
* fix: listAddressLinks now returns all links correctly
* fix: linkAddress now sends link to address-server


oed:
Fixing tests needed (should be simple).

PR #575, opened 2019-08-22
fix: multi tab ipfs instances for pubsub

zachferland:


zachferland:
@oed  this can be reviewed, but looking at it again, might have to think about how unused ipfs config/key repos are clean up, right now this specifically cleared indexeddb but wouldn't remove folder for example if used through node, going to see if can use same leveldown interface to delete

src/3box.js
```diff
@@ -723,6 +725,25 @@ class Box {
...
+function initIPFSRepo () {
+  const sessionID = utils.randInt(10000)
+  const ipfsRootPath = '/ipfs/root/' + sessionID
+
+  const levelInstance = new LevelStore(ipfsRootPath)
+  function randLevelStore () { return levelInstance }
+
+  const repo = new IPFSRepo('/ipfs', {
+    storageBackends: {
+      root: randLevelStore
```
oed: If we have a random repo on each start wouldn't the data not be availible? I.e. we get the head from the orbit cache, then we would get the rest of the data from ipfs, but since we are using a new repo the entries would not be available? Am I missing something? Or would it be possible to only randomize the ipfs keys?

src/3box.js
```diff
@@ -731,13 +752,23 @@ async function initIPFS (ipfs, iframeStore, ipfsOptions) {
...
+      ipfs = new IPFS(ipfsOptions)
       ipfs.on('error', error => {
         console.error(error)
         reject(error)
       })
-      ipfs.on('ready', () => resolve(ipfs))
+      ipfs.on('ready', () => {
+        resolve(ipfs)
+        if (ipfsRepo && window && window.indexedDB) {
+          window.indexedDB.deleteDatabase(ipfsRepo.rootPath)
```
oed: I'm don't understand this. Why do we delete the db right after ipfs in initialized? Wouldn't the db be used in this session?

src/3box.js
zachferland: yeah this is just the root portion of the repo, which includes config info and in the case of ipfsjs still includes the keys/id. The rest of the repo still persists, so things likes blocks (all the data) 

but maybe there is a way to just create a  new id/keys instead of to the same effect creating a new folder/deleting every time

src/3box.js
zachferland: yeah the db is used, this is blocked until the process (ipfs repo) lets go of its lock (or open status) then it is deleted. So it still get deleted even after the window is abruptly closed

zachferland: This works nice for indexeddb (browser env) but something more stable for node would also handle the filesystem. In that case it would have look to delete older key repos on start instead of on close

zachferland:
Its not the best solution, a bit hacky, but rather simple. But comments above should clear it up a bit. Im gonna check two things 1) if we can just change keys/id in a more predictable way 2) handle node env better

src/3box.js
oed: Aha, makes sense!

src/3box.js
oed: I don't think this would be much of an issue in node though since we don't have multiple tabs :P Maybe just don't randomize when in node?

zachferland: yeah that's one option, probs simplest at moment. But would be nice to get one logical instance of ipfs per instance of 3box, if running tests or multiple instances in one directory. But yeah I think should skip worrying about that for now 👍

src/3box.js
```diff
@@ -723,6 +725,29 @@ class Box {
   }
 }
 
+function initIPFSRepo () {
+  let repoOpts = {}
+  let ipfsRootPath
+
+  // if in browser, create unique root storage, and ipfs id on each instance
```
oed: Maybe add a bit more here so that readers understand why this is done :)

oed:
Tests seem fine now, just one lint issue.

zachferland:
Had to move to ipfs repo 0.26.6, ^0.27.0 seems to both name/prefix dbs differently and initialize differently. Didn't look into exactly what changed there. 

As we continue to further customize our implementation we will have to be careful with versions and version compatibilities for example ipfs-js with ipfs-repo

Now need to further test still.

oed:
Do we actually need to explicitly require ipfs-repo? Isn't it already part of js-ipfs?

oed:
Tried to run this again. Now it's just getting stalled on initIPFS. It seems like the `'ready'` event is never fired. Neither is `'error'`. So the promise never returns.

zachferland:
this might be if you ran with the same cache that you ran this branch with before (when broken)?

oed:
No luck still. Tested in both Firefox and chromium.
Also I get this if I try to install with npm ci:
```sh
 npm ci                                                                                                             18:27
npm WARN prepare removing existing node_modules/ before installation
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR!
npm ERR!
npm ERR! Invalid: lock file's datastore-level@0.11.0 does not satisfy datastore-level@^0.12.1
npm ERR!

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/oed/.npm/_logs/2019-09-09T16_27_28_784Z-debug.log
```

zachferland:
Ok made those dependencies implicit (and to follow ipfs), and tested some more, all set now

src/3box.js
```diff
@@ -23,6 +23,8 @@ const utils = require('./utils/index')
 const idUtils = require('./utils/id')
 const config = require('./config.js')
 const API = require('./api')
+const IPFSRepo = require('ipfs-repo')
+const LevelStore = require('datastore-level/src/index.js')
```
oed: Why are we requiring the file directly?

zachferland: yeah can fix that, think left over from when I reference browser implementation specifically

zachferland:
Going to add release commit now on to this branch and then merge and release

PR #582, opened 2019-09-04
Feat/ghost thread

oed:
This is a Work In Progress PR for the ghost chat, for easier reviewing.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -115,6 +115,32 @@
             <button id="postThread" type="button" class="btn btn btn-primary">Post</button>
           </div>
 
+          <!-- GhostChat Start -->
+          <h3> Ghost Chat: </h3>
+          <button id="joinChatBtn" type="button" class="btn btn btn-primary" >Join chat</button>  </br></br>
+          <button id="leaveChatBtn" type="button" class="btn btn btn-primary" style="display: none;">Leave chat</button>  </br></br>
+
+          <div id="chatMembers" style="display: none;">
+            <h5> Members: </h5>
+            <p>
+              <span id="chatMemberList"></span>
+            </p>
+          </div></br>
+
+          <p>
+            <span id="chatACError"></span>
+          </p>
+
+          <div id="messages" style="display: none;">
+            <h4> Messages: </h4>
+            <p>
+              <span id="chatData"></span>
+            </p>
+            <input type="text" id="messageField" placeholder="Type your message">
+            <button id="postChatBtn" type="button" class="btn btn btn-primary">Post</button>
+          </div>
+          <!-- GhostChat End -->
+
         </div>
       </div>
 
--- a/example/index.html
+++ b/example/index.html
@@ -102,6 +102,8 @@
           <input type="text" id="threadfirstModerator" placeholder="Thread Root Moderator"></br>
             <span > Members Only Thread </span>
           <input type="radio" name="members" id='members' value="true"> </br>
+            <span > Ghost ? </span>
+          <input type="radio" name="ghostCheck" id='ghostCheck' value="true"> </br>
           <button id="joinThread" type="button" class="btn btn btn-primary" >Join thread</button>  </br></br>
 
           <div id="threadModeration" style="display: none;">
@@ -133,34 +135,6 @@
             <input type="text" id="postMsg" placeholder="Type your message">
             <button id="postThread" type="button" class="btn btn btn-primary">Post</button>
           </div>
-
-          <!-- GhostChat Start -->
-          <h3> Ghost Chat: </h3>
-          <button id="joinChatBtn" type="button" class="btn btn btn-primary" >Join chat</button>  </br></br>
-          <button id="leaveChatBtn" type="button" class="btn btn btn-primary" style="display: none;">Leave chat</button>  </br></br>
-          
-          <button id="backlogBtn" type="button" class="btn btn btn-primary" style="display: none;">Get Posts</button>  </br></br>
-
-          <div id="chatMembers" style="display: none;">
-            <h5> Members: </h5>
-            <p>
-              <span id="chatMemberList"></span>
-            </p>
-          </div></br>
-
-          <p>
-            <span id="chatACError"></span>
-          </p>
-
-          <div id="messages" style="display: none;">
-            <h4> Messages: </h4>
-            <p>
-              <span id="chatData"></span>
-            </p>
-            <input type="text" id="messageField" placeholder="Type your message">
-            <button id="postChatBtn" type="button" class="btn btn btn-primary">Post</button>
-          </div>
-          <!-- GhostChat End -->
 
         </div>
       </div>
--- a/example/index.html
+++ b/example/index.html
@@ -13,6 +13,25 @@
 
     <div id="controlls" style="display: none; padding: 5px 100px;">
 
+      <div style="padding: 0px 0px 25px 0px;">
+        <h2> Identity </h2>
+        <h5 style="padding: 20px 0px 10px 0px"> DID: </h5>
+        <p>
+          <span id="didInfo"></span>
+        </p>
+
+        <h3> Address Links </h3>
+        <p>
+          <span id="addressLinks"></span>
+        </p>
+
+        <h5> Current address linked? </h5>
+        <p>
+          <span id="addressLinked"></span>
+        </p>
+        <button id="linkAddress" type="button" class="btn btn-primary">Link address</button>
+      </div>
+
       <!-- Public Store Div -->
       <div style="padding: 0px 0px 25px 0px;">
         <h2> Public Store </h2>
--- a/example/index.html
+++ b/example/index.html
@@ -119,6 +119,8 @@
           <h3> Ghost Chat: </h3>
           <button id="joinChatBtn" type="button" class="btn btn btn-primary" >Join chat</button>  </br></br>
           <button id="leaveChatBtn" type="button" class="btn btn btn-primary" style="display: none;">Leave chat</button>  </br></br>
+          
+          <button id="backlogBtn" type="button" class="btn btn btn-primary" style="display: none;">Get Posts</button>  </br></br>
 
           <div id="chatMembers" style="display: none;">
             <h5> Members: </h5>
--- a/example/index.js
+++ b/example/index.js
@@ -183,11 +183,8 @@ bopen.addEventListener('click', event => {
         const name = 'example 3box dapp'
         const threeId = box._3id
         const ipfs = box._ipfs
-        ipfs.swarm.connect('/ip4/127.0.0.1/tcp/4015/ws/ipfs/QmXGbAwcGmCFsMNMA5NHjz8r16WP9DHPLBf6RNEtN1kNji', console.log)
-        // TODO: solve connection issue
         const chat = box.spaces[window.currentSpace].joinChat()
         window.currenChat = chat
-        console.log(window.currenChat)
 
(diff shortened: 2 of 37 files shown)


src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+  threeIdToPeerId (did) {
+    return this._usersOnline[did]
+  }
+
+  /**
+   * Get backlog of all past messages
+   *
+   * @return    {Array<Object>}      users online
+   */
+  get backlog () {
```
oed: Rename to `getPosts()` to be consistent with the Thread api.

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    return [...this._backlog]
+  }
+
+  // Announce entry in chat and share our 3id and peerID, empty jwt suffices
+  async announce (to) {
```
oed: usually private methods start with a `_`

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * Post a message to the thread
+   *
+   * @param     {Object}    message                 The message
+   * @param     {String}    to                      PeerID to send the message to (optional)
+   * @return    {String}                            The postId of the new post
+   */
+  async post (message, to) {
```
oed: Why do we have a to option here? It's used internally in the backlog sending? Don't think that needs to be part of the public method :)

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    !to ? await this.broadcast({ type: 'chat', message })
+    : await this.sendDirect(to, { type: 'chat', message })
+  }
+
+  /**
+   * Request a backlog of past messages from peers in the chat
+   *
+   */
+  async requestBacklog () {
```
oed: This should be a private method.

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    await this.broadcast({ type: 'request_backlog' })
+  }
+
+  /**
+   * Leave the chat
+   *
+   */
+  async leaveChat () {
```
oed: I think just `close()` would be a better name here :)

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    // await this.broadcast({ type: 'leave' })
+    await this._room.leave()
+  }
+
+  /**
+   * Broadcast a message to peers in the room
+   *
+   * @param     {Object}    message                 The message
+   */
+  async broadcast (message) {
```
oed: This should be a private method.

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    this._room.broadcast(jwt)
+  }
+
+  /**
+   * Send a direct message to a peer
+   *
+   * @param     {String}    peerID              The PeerID of the receiver
+   * @param     {Object}    message             The message
+   */
+  async sendDirect (to, message) {
```
oed: This should also be private I think. And since this method exists I don't think we need a `to` in the `post` method.

src/ghost.js
```diff
@@ -0,0 +1,186 @@
...
+    this._spaceName = name.split('.')[2]
+    this._3id = threeId
+    this._room = Room(ipfs, name) // instance of ipfs pubsub room
+    this._peerId = ipfs._peerInfo.id.toB58String()
+
+    this._usersOnline = {}
+    this._backlog = new Set() // set of past messages
+
+    this._room.on('message', async ({ from, data }) => {
+      // const { payload, issuer } = await this._verifyData(data)
```
oed: Did you manage to solve this?

src/ghost.js
```diff
@@ -0,0 +1,186 @@
+const EventEmitter = require('events').EventEmitter
+const { decodeJWT, verifyJWT } = require('did-jwt')
+const Room = require('ipfs-pubsub-room')
+
+class GhostChat extends EventEmitter {
+
+  // TODO: map did to peer id
+  // TODO: for online list, what do we do about late joiners who don't know who's online yet
```
oed: I'm pretty sure we could use `ipfs.pubsub.ls()` to get a list of peerIDs. Then we could send request to each for getting their 3IDs. We can also use the same command to see if any user has left without calling the leaveChat function.

src/ghost.js
oed: Feels like this should be called automatically when joining a chat.

src/ghost.js
ghiliweld: Yes I did

src/ghost.js
ghiliweld: I made `sendDirect` private but I do think think `post` should be kept how it is, it provides a public way to send both broadcasted and direct messages.

src/ghost.js
ghiliweld: This is what I did to solve it

```js
this._room.on('peer joined', (peer) => this.announce(peer))
```

Once a peer joins the room, our peer immediately announces its presence to the peer.

ghiliweld:
This how I'm going about using `onUpdate` and `onNewCapabilities` in the ghost api.

```js
onUpdate (updateFn) {
  this.on('message', updateFn)
}

onNewCapabilites (updateFn) {
  this.on('user-joined', updateFn)
  this.on('user-left', updateFn)
}
```

As for how `user-joined` and `user-left` are emitted, it goes like this.

```js
this.emit('user-joined', 'joined', did, peerID)
this.emit('user-left', 'left', did, peerID)
```
A `joined/left` string is emitted so that the `updateFn` in `onNewCapabilities` can detect which type of event it is.

src/space.js
```diff
@@ -79,6 +81,11 @@ class Space {
     return thread
   }
 
+  joinChat () {
```
oed: Instead of having a `space.joinChat()` method we should use the existing `space.joinThread()` method. This would make the API the same for developers that are already familiar with threads.

To join a ghostChat you could simply call:
```js
const thread = await space.joinThread('name', { ghost: true })
```

oed: We should also add this to the jsdoc on `joinThread`.

oed:
Make the `requestBacklog` method called after peers join so that we can get backlog by only joining the chat.

src/space.js
```diff
@@ -1,10 +1,12 @@
 const KeyValueStore = require('./keyValueStore')
 const Thread = require('./thread')
+const GhostChat = require('./ghost')
 const { throwIfUndefined, throwIfNotEqualLenArrays } = require('./utils')
 const OrbitDBAddress = require('orbit-db/src/orbit-db-address')
 
 const nameToSpaceName = name => `3box.space.${name}.keyvalue`
 const namesTothreadName = (spaceName, threadName) => `3box.thread.${spaceName}.${threadName}`
+const namesToChatName = spaceName => `3box.chat.${spaceName}`
```
oed: Should be `3box.ghost.${spaceName}.${chatName}`

src/space.js
```diff
@@ -65,18 +67,25 @@ class Space {
    * @param     {String}    opts.firstModerator     DID of first moderator of a thread, by default, user is first moderator
    * @param     {Boolean}   opts.members            join a members only thread, which only members can post in, defaults to open thread
    * @param     {Boolean}   opts.noAutoSub          Disable auto subscription to the thread when posting to it (default false)
+   * @param     {Boolean}   opts.ghost              Enable ephemeral messaging via Ghost Thread (optional)
    *
-   * @return    {Thread}                            An instance of the thread class for the joined thread
+   * @return    {Thread/GhostChat}                  An instance of the thread class for the joined thread
```
oed: This should still say only Thread

src/space.js
```diff
@@ -65,18 +67,25 @@ class Space {
...
-    const subscribeFn = opts.noAutoSub ? () => {} : this.subscribeThread.bind(this)
-    if (!opts.firstModerator) opts.firstModerator = this._3id.getSubDID(this._name)
-    const thread = new Thread(namesTothreadName(this._name, name), this._replicator, this._3id, opts.members, opts.firstModerator, subscribeFn)
-    const address = await thread._getThreadAddress()
-    if (this._activeThreads[address]) return this._activeThreads[address]
-    await thread._load()
-    this._activeThreads[address] = thread
-    return thread
+    if (opts.ghost) {
+      const chat = new GhostChat(namesToChatName(name), this._replicator, this._3id)
```
oed: This line is missing the space name in the `namesToChatName` function.

oed: I would suggest the following change as well:
```suggestion
      const ghostAddress = namesToChatName(this._name, name)
      if (!this._activeThreads[ghostAddress]) {
          this._activeThreads[ghostAddress] = new GhostChat(ghostAddress, this._replicator, this._3id)
      }
      return this._activeThreads[ghostAddress]
```

src/ghost.js
```diff
@@ -0,0 +1,212 @@
+const EventEmitter = require('events').EventEmitter
+const { verifyJWT } = require('did-jwt')
+const Room = require('ipfs-pubsub-room')
+
+class GhostChat extends EventEmitter {
+
+  /**
+   * Please use **space.joinChat** to get the instance of this class
+   */
```
oed: We can just remove this now :)

src/ghost.js
```diff
@@ -0,0 +1,212 @@
+const EventEmitter = require('events').EventEmitter
+const { verifyJWT } = require('did-jwt')
+const Room = require('ipfs-pubsub-room')
+
+class GhostChat extends EventEmitter {
```
oed: Maybe we should rename from GhostChat to GhostThread?

oed:
Also if you could do `$ git merge develop` the conflicts should be gone :)

src/3box.js
```diff
@@ -510,10 +510,18 @@ class Box {
...
+    const data = (await this._ipfs.dag.put(proof)).toBaseEncodedString()
+    await this._ipfs.pin.add(data)
+    utils.fetchJson(this._serverUrl + '/link', proof).catch(console.error)
+    const linkExist = await this._linkCIDExists(data)
+=======
   async _writeRootstoreEntry (type, payload) {
     const cid = (await this._ipfs.dag.put(payload)).toBaseEncodedString()
     await this._ipfs.pin.add(cid)
     const linkExist = await this._typeCIDExists(type, cid)
+>>>>>>> 058dde8ba8b887d5d1839b216f55c6097a7ed01d
```
oed: Fix this.

example/index.html
```diff
@@ -134,6 +134,34 @@ <h4> Posts: </h4>
             <button id="postThread" type="button" class="btn btn btn-primary">Post</button>
           </div>
 
+          <!-- GhostChat Start -->
```
oed: Did you have a chance to look at if this can be removed and just add an additional radiobutton to the threads code?

ghiliweld: Yeah that'll be finished today

src/3box.js
ghiliweld: Fixed

oed:
Looks like there is a merge conflict in package-lock.json. You can fix it by merging develop into your branch.
The linter is still failing. You can run `npm run lint` locally and fix the errors.

example/index.js
```diff
@@ -153,25 +160,93 @@ bopen.addEventListener('click', event => {
...
           })
-        })
+        }
       }
 
       postThread.addEventListener('click', () => {
         window.currentThread.post(postMsg.value).catch(updateThreadError)
       })
 
+      // Ghost Chat Start
```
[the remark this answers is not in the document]
ghiliweld: Right, I forgot to remove it

PR #597, opened 2019-09-17
Link Smart Accounts

jsidorenko:


src/3box.js
```diff
@@ -17,7 +17,8 @@ const LevelStore = require('datastore-level')
 
 const ACCOUNT_TYPES = {
   ethereum: 'ethereum',
-  ethereumEOA: 'ethereum-eoa'
+  ethereumEOA: 'ethereum-eoa',
+  ethereumContract: 'ethereum-contract'
```
oed: 👍 , or maybe this should be called `ERC1271`?

jsidorenko: I would stick to some generic naming, as there are no confirmed standard for this yet. We could extend the validation functions later, but the account type will still be the Smart Contract based accound

src/3box.js
```diff
@@ -540,6 +568,14 @@ class Box {
     return links.find(link => link.address.toLowerCase() === address)
   }
 
+  async _detectAddressType (address) {
+    const bytecode = await utils.getCode(this._web3provider, address).catch(() => null)
```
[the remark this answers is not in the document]
jsidorenko: `try..catch` is a heavy thing which compilers usually skip during the optimisation phase. Probably it's fine to use them in web/node env, although as for me the `try..catch` looks less readable comparing to `.catch()`

jsidorenko:
> It looks like you assume that we'd use an EOA account to sign a message that would later be validated by the `ERC1271` method. Is that right?
> 
> Let me know if you are thinking about this differently!

Right, in Abridged sdk case there is an EOA account that controls the smart account. So, EOA could sign a message in a standard way and smart contract will be able to validate that message later

jsidorenko:
> As I understand it each smart contract wallet can encode these signatures differently. Some might only require one signature while others require more. 
In that case, we could accept the passed as a param signature and store it 'as is' + create the validator method

src/utils/index.js
```diff
@@ -76,11 +79,12 @@ module.exports = {
...
   },
 
-  getLinkConsent: async (fromAddress, toDID, ethereum) => {
+  getLinkConsent: async (fromAddress, toDID, ethereum, contractAddress) => {
     const timestamp = Math.floor(new Date().getTime() / 1000)
-    const text = getMessageConsent(toDID, timestamp)
+    const text = getMessageConsent(toDID, timestamp, contractAddress)
     const msg = '0x' + Buffer.from(text, 'utf8').toString('hex')
     const params = [msg, fromAddress]
+    if (contractAddress) params.push(contractAddress)
```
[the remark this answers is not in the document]
jsidorenko: You can't sign anything with the smart contract as it has no private key, but you can sign the data with your EOA on behalf of your smart contract. Again, this depends on the software we use to manage the smart account. We could get rid of most of those hacks if they hack the `personal_sign` method by using the private key of EOA on the background

src/3box.js
oed: The main reason I would prefer `ERC1271` here would be that if another standard emerges there would be an easy way to distinguish between links made in this way and links made with a new standard :)

jsidorenko: You're right!
How about putting that info into the `linkData`?

src/utils/index.js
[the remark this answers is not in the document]
oed: Right, so we might be trying to solve two slightly different use cases here. 
1. I want to link a contract wallet to an existing 3box
1. I want to automatically link my account to a 3box

In the case of 1. we could simply provide a proof object to the `linkAccount` method (should already be supported)

Case 2. happens when there is a new 3Box account and you first set data. Then `_linkProfile` is automatically called (you can search the code for when this happens). If this is done by a smart contract based wallet we need to detect that and format the stored signature correctly.

Hope that clarifies some things.

src/3box.js
oed: This would end up in the `type` property as it is now right?

jsidorenko: :D right
Ok, so we can either create a sub-type, or just rename this type into smth else. You choose ;)
btw, in your spec `3ip-5` you wrote that the value should be `ethereum-contract`, so this spec should be changed as well

src/utils/index.js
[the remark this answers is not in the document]
jsidorenko: Got it, I will change the code accordingly tomorrow

src/3box.js
oed: Yeah, let's update it in the `3ip-5` spec and change this type :)

oed:
> Right, in Abridged sdk case there is an EOA account that controls the smart account. So, EOA could sign a message in a standard way and smart contract will be able to validate that message later

Makes sense, but I assume the abridged SDK will provide an `window.ethereum` provider that dapp devs could use? Or will that be abstracted away?

> In that case, we could accept the passed as a param signature and store it 'as is' + create the validator method

Which passed param are you refering to here?

jsidorenko:
> Makes sense, but I assume the abridged SDK will provide an window.ethereum provider that dapp devs could use? Or will that be abstracted away?

If only I knew :D we're integrating them into the mobile app, there is no such thing as window.ethereum

jsidorenko:
> Which passed param are you refering to here?

`link.proof` param in the `linkAddress()` method

oed:
> link.proof param in the linkAddress() method

Ok, then yes makes sense! :)

src/3box.js
```diff
@@ -526,12 +540,17 @@ class Box {
...
   async _readAddressLinks () {
     const links = await this.replicator.getAddressLinks()
-    return Promise.all(links.map(async linkObj => {
+    const allLinks = await Promise.all(links.map(async linkObj => {
       if (!linkObj.address) {
-        linkObj.address = await utils.recoverPersonalSign(linkObj.message, linkObj.signature)
+        linkObj.address = await utils.recoverPersonalSign(linkObj.message, linkObj.signature).catch(() => null)
+      }
+      const isErc1271 = linkObj.type === ACCOUNT_TYPES.erc1271
+      if (!(await utils.isValidSignature(linkObj, isErc1271, this._web3provider))) {
```
oed: Makes sense.

src/3box.js
```diff
@@ -448,28 +454,50 @@ class Box {
...
+          timestamp: consent.timestamp,
+          signature: consent.sig
+        }
+        : {
+          version: 1,
+          type: ACCOUNT_TYPES.ethereumEOA,
+          message: consent.msg,
+          signature: consent.sig,
+          timestamp: consent.timestamp
+        }
```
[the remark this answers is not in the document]
oed: Did you see this comment before? :)

src/utils/index.js
```diff
@@ -97,6 +100,45 @@ module.exports = {
...
+      id: 1,
+      method,
+      params
+    })
+    return code
+  },
+
+  isValidSignature: async (linkObj, isErc1271, web3Provider) => {
+    if (!linkObj.address) return false
+    if (!isErc1271) return true
```
oed: Why always return true here?

src/3box.js
```diff
@@ -450,6 +451,7 @@ class Box {
 
   async _linkProfile () {
     const address = await this._3id.getAddress()
+    const addressType = await this._detectAddressType(address)
```
oed: We can move this to around line 466 since it's only needed in that block!

src/3box.js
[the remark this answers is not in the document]
jsidorenko: Sure, already changed ;)

src/utils/index.js
jsidorenko: The only check we're doing for EOA is the address extraction before we call the `isValidSignature`.  Maybe we could rewrite this into:
```javascript
if (!isErc1271) {
  if (!linkObj.address) return false
  return true
}
```

jsidorenko: We could event extract this into `isValidEOALink` and `isValid1271Link` and check the input params in both methods

src/utils/index.js
```diff
@@ -97,6 +98,46 @@ module.exports = {
...
+
+    const abi = [
+      'function isValidSignature(bytes memory _messageHash, bytes memory _signature) public view returns (bytes4 magicValue)'
+    ]
+    const ethersProvider = new ethers.providers.Web3Provider(web3Provider)
+    const contract = new ethers.Contract(linkObj.address, abi, ethersProvider)
+    const returnValue = await contract.isValidSignature(linkObj.message, linkObj.signature)
+
+    return returnValue === MAGIC_ERC1271_VALUE
```
oed: Oh neat, this was simpler than I imagined :)

oed:
Looks great now! Would you mind squashing the commits? Then I'll merge this.

PR #605, opened 2019-09-24
Release/1.11.0

oed:
# Release Notes

## v1.11.0 - 2019-09-24
* feat: 3box instance can now be created with an instance of the IdentityWallet
* feat: smarter data syncing logic
* feat: support erc1271 address links (your contract wallet can now have a 3Box!)
* feat: verify link proofs added by the `linkAddress` method

* fix: sync rootstore before resolving promise from openBox

### Updated data sync api
To wait for the data to be fully synced from the network you can now simply await these publicly exposed promises:
* `await box.syncDone` - for the main data storage
* `await space.syncDone` - for data in the given space

### DIFF
--- a/src/keyValueStore.js
+++ b/src/keyValueStore.js
@@ -172,7 +172,7 @@ class KeyValueStore {
   }
 
   async _load (odbAddress) {
-    const key = this._3id.getKeyringBySpaceName(this._name).getPublicKeys(true).signingKey
+    const key = (await this._3id.getPublicKeys(this._space, true)).signingKey
     const opts = {
       ...ORBITDB_OPTS,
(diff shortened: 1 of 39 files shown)


src/keyValueStore.js
```diff
@@ -238,7 +202,7 @@ class KeyValueStore {
    *
    * @return    {Array<Object>}     Array of ordered log entry objects
    */
-  get log () {
+  async log () {
```
zachferland: why async now?

oed: Because the decrypt function in the IdentityWallet has to be async. If this function was sync then the api would be different for private data.

PR #613, opened 2019-10-01
Adding a 12 hour backlog limit to GhostThreads

ghiliweld:


src/ghost.js
```diff
@@ -68,7 +68,16 @@ class GhostThread extends EventEmitter {
    * @return    {Array<Object>}      users online
    */
   async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+    const _12hrInSecs = 12 * 60 * 60
```
oed: Move this constant to the top of the file. Usually constants are named with all capital letters. 
I would name it someting like: 
```js
const POST_AGE_LIMIT = 12 * 60 * 60 // 12 hours
```

src/ghost.js
```diff
@@ -68,7 +68,16 @@ class GhostThread extends EventEmitter {
...
+    const posts = [...this._backlog]
+      .map(msg => JSON.parse(msg))
+      .sort((p1, p2) => p1.timestamp - p2.timestamp)
+      .filter(msg => msg.timestamp > (Math.floor(Date.now() / 1000) - _12hrInSecs))
+
+    posts.forEach(msg => {
+      this._backlog.add(JSON.stringify(msg))
```
oed: Not sure why you are doing this? It seems like these *msg* would already be in the `this._backlog` set?

src/ghost.js
```diff
@@ -68,7 +68,16 @@ class GhostThread extends EventEmitter {
    * @return    {Array<Object>}      users online
    */
   async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+    const _12hrInSecs = 12 * 60 * 60
+    const posts = [...this._backlog]
+      .map(msg => JSON.parse(msg))
+      .sort((p1, p2) => p1.timestamp - p2.timestamp)
+      .filter(msg => msg.timestamp > (Math.floor(Date.now() / 1000) - _12hrInSecs))
```
oed: Should we add this filter to before we send the backlog to other peers as well?

src/ghost.js
ghiliweld: The idea was that if we keep a shorter version of the backlog it would make looping through it more efficient. Once we get a limited version of it we should just keep it on hand instead of always looping through the whole history.

src/ghost.js
ghiliweld: When we send the backlog to other peers, we go use the `getPosts()` method as well, so we are sending them a filtered version.

```js
case 'request_backlog':
this.getPosts()
 .then(posts => this._sendDirect({ type: 'backlog_response', message: posts }, from))
 break
```

src/ghost.js
oed: Ok, but then you need to clear out the `this._backlog` variable before right? Otherwise it would still contain all messages!

src/ghost.js
oed: Oh, sorry totally missed that!

PR #615, opened 2019-10-02
GhostThread: Backlog limit added

ghiliweld:
This PR got rid of the other bad commits, should be all clear as well.

### DIFF
--- a/src/ghost.js
+++ b/src/ghost.js
@@ -2,6 +2,8 @@ const EventEmitter = require('events').EventEmitter
 const { verifyJWT } = require('did-jwt')
 const Room = require('ipfs-pubsub-room')
 
+const DEFAULT_BACKLOG_LIMIT = 50
+
 class GhostThread extends EventEmitter {
   constructor (name, { ipfs }, threeId, opts = {}) {
     super()
@@ -13,7 +15,7 @@ class GhostThread extends EventEmitter {
 
     this._members = {}
     this._backlog = new Set() // set of past messages
-    this._backlogLimit = opts.backlogLimit || 50 // 50 message default backlog limit
+    this._backlogLimit = opts.ghostBacklogLimit || DEFAULT_BACKLOG_LIMIT
 
     this._room.on('message', async ({ from, data }) => {
       const { payload, issuer } = await this._verifyData(data)
--- a/src/ghost.js
+++ b/src/ghost.js
@@ -68,7 +68,16 @@ class GhostThread extends EventEmitter {
    * @return    {Array<Object>}      users online
    */
   async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+    const _12hrInSecs = 12 * 60 * 60
+    const posts = [...this._backlog]
+      .map(msg => JSON.parse(msg))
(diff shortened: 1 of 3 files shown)


src/ghost.js
```diff
@@ -68,7 +68,16 @@ class GhostThread extends EventEmitter {
...
   async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+    const _12hrInSecs = 12 * 60 * 60
+    const posts = [...this._backlog]
+      .map(msg => JSON.parse(msg))
+      .sort((p1, p2) => p1.timestamp - p2.timestamp)
+      .filter(msg => msg.timestamp > (Math.floor(Date.now() / 1000) - _12hrInSecs))
+
+    posts.forEach(msg => {
+      this._backlog.add(JSON.stringify(msg))
```
oed: I'm pretty sure you need to do something like `this._backlog = new Set()` before you add these messages. Otherwise the old messages would be persisted.

ghiliweld:
@oed I also added the code for getting that last n posts in `getPosts()` please review that.

src/ghost.js
```diff
@@ -67,8 +69,26 @@ class GhostThread extends EventEmitter {
...
+      const posts = sorted.slice(-num)
+      posts.forEach(msg => {
+        this._backlog.add(JSON.stringify(msg))
+      })
+
+      return posts
+    } else {
+      const posts = sorted.filter(msg => msg.timestamp > (Math.floor(Date.now() / 1000) - POST_AGE_LIMIT))
+      posts.forEach(msg => {
+        this._backlog.add(JSON.stringify(msg))
```
oed: Still wondering about this. `this._backlog` will already contain all messages being added to it here.

src/ghost.js
```diff
@@ -67,8 +69,26 @@ class GhostThread extends EventEmitter {
    *
    * @return    {Array<Object>}      users online
    */
-  async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+  async getPosts (num = null) {
```
oed: Rather than having num passed as a parameter here I think we should put it as an optional parameter in the constructor and make it global. Something like `this._backlogLimit`. The main reason we want this is to only send this number of messages to other peers when they join.

oed: Use 50 as default.

For example you could call `space.joinThread('name', { ghost: true, backlogLimit: 123 })`

src/ghost.js
```diff
@@ -67,8 +68,16 @@ class GhostThread extends EventEmitter {
...
-  async getPosts () {
-    return [...this._backlog].map(msg => JSON.parse(msg)).sort((p1, p2) => p1.timestamp - p2.timestamp)
+  async getPosts (num = this._backlogLimit) {
+    const posts = [...this._backlog]
+      .map(msg => JSON.parse(msg))
+      .sort((p1, p2) => p1.timestamp - p2.timestamp)
+      .slice(-num)
+
+    const stringifiedPosts = posts.map(msg => JSON.stringify(msg))
+    this._backlog = new Set(stringifiedPosts)
```
oed: 👍

PR #616, opened 2019-10-02
GhostThread: added prod signalling servers

ghiliweld:
Fixed bad commits

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -16,7 +16,7 @@ module.exports = {
       Bootstrap: [],
       Addresses: {
         Swarm: [
-          '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
+          '/dns4/p2p.3box.io/tcp/9090/wss/p2p-websocket-star/'
         ]
       }
     }


ghiliweld:
@oed idWallet tests fail here. Any idea why?

oed:
I believe there might be some concurrency issues. I reran the test on circleci and it passed without any modification.

oed:
Alright, I did a bunch of performance testing of this and come up with the following:

* Adding **one** signaling server adds around 3 sec load time to `openBox`.
* Adding **both** signaling servers adds around 6 sec
* The webrtc signaling service can cause ipfs to throw an error on start if you have a plugin that blocks webrtc (I'm using https://github.com/ChrisAntaki/disable-webrtc-firefox)

@zachferland suggested that we connect to the signaling servers using `ipfs.swarm.connect` similar to how we connect to the pinning-node. However this is not possible. I created an issue to ask about it:
https://github.com/ipfs/js-ipfs/issues/2508

The result of the performance tests can be seen here:
https://docs.google.com/spreadsheets/d/1YO0vHFfG0kxOOYEm3xLaWpQ4INicittQIXwCwWRj15w/edit?usp=sharing

## What next?
I think we should comment out the webrtc signaling server for now and just rely on the websocket one.

PR #619, opened 2019-10-06
feat: support for 3ID Provider

oed:
First go at 3id provider support.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,6 @@
 coverage/
 tmp/
 example/
+orbitdb/
+.circleci/
+keystore/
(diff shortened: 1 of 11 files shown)


example/index.js
```diff
@@ -5,7 +5,10 @@ const syncComplete = (res) => {
 bopen.addEventListener('click', event => {
 
   window.ethereum.enable().then(addresses => {
-    window.Box.openBox(addresses[0],  window.ethereum, {}).then(box => {
+    const idw = new IdentityWallet({ seed: '0x9827348745987235978347853475347345768349872384'})
```
zachferland: may make example confusing to have this as default

oed: Thanks, this should be reverted!

PR #632, opened 2019-10-21
feat: return all seen messages in getPosts()

ghiliweld:


src/ghost.js
```diff
@@ -70,15 +70,12 @@ class GhostThread extends EventEmitter {
    *
    * @return    {Array<Object>}      users online
    */
-  async getPosts (num = this._backlogLimit) {
+  async getPosts (num = 0) {
```
oed: Neat

PR #634, opened 2019-10-22
feat: introduced filters

ghiliweld:


src/ghost.js
```diff
@@ -17,9 +17,16 @@ class GhostThread extends EventEmitter {
...
 
+    this._filters = opts.ghostFilters || []
+
     this._room.on('message', async ({ from, data }) => {
       const { payload, issuer } = await this._verifyData(data)
-      if (payload) {
+
+      // looping thru _filters array w map should return an array of booleans
+      // reduce that boolean array to one boolean that indicates whether we pass all the filters
+      const passesFilters = this._filters.map(filter => filter({ payload, issuer, from })).reduce((sum, next) => sum && next, true)
```
oed: Can't we just make the filter function take three params instead of one? I think this would make it a little bit easier to implement filters for developers. 

Btw, you can also simplify this line by removing the `.map`.
Would look something like: 
```js
const passesFilters = this._filters.reduce((acc, filter) => acc && filter({ payload, issuer, from }), true)
```
If we change the filter params as described above:
```js
const passesFilters = this._filters.reduce((acc, filter) => acc && filter(payload, issuer, from), true)
```

src/__tests__/ghost.test.js
```diff
@@ -27,7 +27,7 @@ describe('Ghost Chat', () => {
   // jest.setTimeout(2000000)
 
   beforeAll(async () => {
-    ipfs = await utils.initIPFS(11);
+    ipfs = await utils.initIPFS(4);
```
oed: Why did you change this? I'm pretty sure that 4 & 5 is being used by another test.

src/ghost.js
```diff
@@ -17,9 +17,16 @@ class GhostThread extends EventEmitter {
...
     this._backlogLimit = opts.ghostBacklogLimit || DEFAULT_BACKLOG_LIMIT
 
+    this._filters = opts.ghostFilters || []
+
     this._room.on('message', async ({ from, data }) => {
       const { payload, issuer } = await this._verifyData(data)
-      if (payload) {
+
+      // looping thru _filters array w map should return an array of booleans
+      // reduce that boolean array to one boolean that indicates whether we pass all the filters
```
oed: This comment is out of date :)

src/__tests__/ghost.test.js
ghiliweld: Oh it just didn't seem right cause at the end of the test we close 4 and 5. I'll change it back

oed: aha! Change the numbers in the end of the test instead then :D

PR #652, opened 2019-11-17
update to ethers v5

ghiliweld:


ghiliweld:
Thank you for that @oed was just about to ask

oed:
Before merging I'd like for @zachferland to have a quick look.

PR #657, opened 2019-11-20
Bugfix: eth_getCode add second parameter "latest"

PeterTheOne:
Without the second parameter infura `Geth/v1.8.27-omnibus-c94f741c/linux-amd64/go1.11.1` returns an error:
```
"code": -32602,
"message": "missing value for required argument 1"
```
Parity seems to have no problem with the missing parameter.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -122,7 +122,7 @@ module.exports = {
 
   getCode: async (ethereumProvider, address) => {
     const method = 'eth_getCode'
-    const params = [address]
+    const params = [address, 'latest']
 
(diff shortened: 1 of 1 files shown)


PeterTheOne:
Looks like geth behaves like this for all these function https://github.com/ethereum/wiki/wiki/JSON-RPC#the-default-block-parameter

oed:
Makes sense. Ty!

oed:
FYI: changed the target branch to develop.

PeterTheOne:
Ok, looking forward to a merge :)

oed:
This should get released sometime this week I think!

Issue #640, opened 2019-10-31
New API for creating a 3box instance


oed:
Right now 3Box is gets instantiated at the same time as the user is authenticated. We have been exploring a new way of creating a 3Box instance that would allow developers to join threads and sync data without authenticating the users. Users can then be authenticated later when needed. Below is a first draft of how this protocol could look like:

### Box.create
```js
const box = await Box.create(provider, opts)
```
Creates an instance of 3box. Does not do any authentication. Possibly has a cache of last synced user and opens the DB in the background since it's likely that this db will be opened again.

- `provider` - the json-rpc provider to be used
- `opts` - optional parameters

### box.joinThread
```js
const thread = box.joinThread(space, name, opts)
```
Join a thread without being logged in.

- `space` - the name of the space this thread is in
- `name` - the name of the thread
- `opts` - optional parameters

### box.auth
```js
await box.auth(spaces, opts)
```
Authenticate the dapp to access the users data.

- `spaces` - an array of space names to be accessed
- `opts` - optional parameters

oed:
Already implemented

PR #670, opened 2019-12-06
Feat/new init api

oed:
This PR adds new apis as specified in https://github.com/3box/3box-js/issues/640

To be merged after #667

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -149,7 +149,9 @@ class Box extends BoxApi {
     }
     // make sure we are authenticated to threads
     spaces.forEach(space => {
-      this.spaces[space]._authThreads(this._3id)
+      if (this.spaces[space]) {
+        this.spaces[space]._authThreads(this._3id)
+      }
     })
   }
 
@@ -221,7 +223,7 @@ class Box extends BoxApi {
    *
    * @return    {Thread}                  An instance of the thread class for the joined thread
    */
-  async joinThread(space, name, opts) {
+  async joinThread (space, name, opts) {
     if (!this.spaces[space]) {
       this.spaces[space] = new Space(space, this.replicator)
     }
--- a/src/3box.js
+++ b/src/3box.js
@@ -47,10 +47,9 @@ class Box extends BoxApi {
    * Please use the **openBox** method to instantiate a 3Box
    * @constructor
    */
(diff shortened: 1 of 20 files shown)


src/3box.js
```diff
@@ -167,6 +208,28 @@ class Box extends BoxApi {
...
+   * @param     {String}    opts.firstModerator     DID of first moderator of a thread, by default, user is first moderator
+   * @param     {Boolean}   opts.members            join a members only thread, which only members can post in, defaults to open thread
+   * @param     {Boolean}   opts.noAutoSub          Disable auto subscription to the thread when posting to it (default false)
+   * @param     {Boolean}   opts.ghost              Enable ephemeral messaging via Ghost Thread
+   * @param     {Number}    opts.ghostBacklogLimit  The number of posts to maintain in the ghost backlog
+   * @param     {Array<Function>} opts.ghostFilters Array of functions for filtering messages
+   *
+   * @return    {Thread}                  An instance of the thread class for the joined thread
+   */
+  async joinThread (space, name, opts) {
```
zachferland: seems confusing to have this func here with the same name, considered have openThread and joinThread or still in a space like space.openThread? but i guess if in space then space with need an open/auth flow, so maybe here but  openThread

src/3box.js
```diff
@@ -221,6 +284,7 @@ class Box extends BoxApi {
    * @property {String} DID        the DID of the user
    */
   get DID () {
+    if (!this._3id) throw new Error('DID: auth required')
```
zachferland: could probs add this constraint to some func/class modifier/wrap, but probably clearer and simpler to keep as is for now

oed: yeah, effort vs gain :)

src/3box.js
oed: So just rename this fn to `openThread`? The thing that might become confusing is to have two different names for two methods that essentially does the same. Note that if you join a thread from here and then call `box.auth` for the given space you will automatically be able to post in this thread.

oed: On second thought, `openThread` here makes alot of sense. Going to change it!

oed:
Thanks @zachferland I'd say wait to run tests until we prepare the release branch :)

PR #677, opened 2019-12-11
Release/1.15.0

oed:
## TODO:
- [ ] manually test old and new initialization interface
- [ ] release updated documentation (can happen right after the code is released)
```
This release is scheduled for 2019-12-13
```
# Release Notes

## v1.15.0 - 2019-12-13
This release features a new interface for how to create and authenticate to a 3Box, it also adds the ability to open a thread before being authenticated to a space.

---
* feat: new initialization interface
* fix: add postId to ghost chat messages
* fix: don't allow space names with dots


zachferland:
👍 

did some quick manual testing, looks good. Just looks like from clear cache the time to return the box is a lot slower, not the time between box return and syncing (but maybe same as last few releases), just original box return. 

not needed for this release, but in future would make sense to update example to use this interface

PR #678, opened 2019-12-12
feat: 3id migration

zachferland:
@oed already did all the hard work with recent refactors and work :)

~~just pushing up quickly, but need to double check things and probably fix some tests~~

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -15,6 +15,7 @@ const config = require('./config.js')
 const BoxApi = require('./api')
 const IPFSRepo = require('ipfs-repo')
 const LevelStore = require('datastore-level')
+const didJWT = require('did-jwt')
 
 const PINNING_NODE = config.pinning_node
 const ADDRESS_SERVER_URL = config.address_server_url
@@ -244,7 +245,7 @@ class Box extends BoxApi {
 
   async _publishRootStore (rootStoreAddress) {
     // Sign rootstoreAddress
-    const addressToken = await this._3id.signJWT({ rootStoreAddress }, { use3ID: true })
+    const addressToken = await this._3id.signJWT({ rootStoreAddress })
     // Store odbAddress on 3box-address-server
     const publish = async token => {
       try {
@@ -319,7 +320,7 @@ class Box extends BoxApi {
       type: 'delete-address-link'
     }
     const oneHour = 60 * 60
-    const deleteToken = await this._3id.signJWT(payload, { expiresIn: oneHour, use3ID: true })
+    const deleteToken = await this._3id.signJWT(payload, { expiresIn: oneHour })
 
     try {
       await utils.fetchJson(this._serverUrl + '/linkdelete', {
@@ -416,7 +417,16 @@ class Box extends BoxApi {
     }
     // Ensure we self-published our did
     // TODO - is this still needed?
-    if (!(await this.public.get('proof_did'))) {
+    const proofdid = await this.public.get('proof_did')
+
+    if (proofdid) {
+      // if prior muport, re publish with 3id including muport
+      const issuer = didJWT.decodeJWT(proofdid).payload.iss
+      if (issuer.includes('muport')) {
+        const jwt = { muport: proofdid }
+        await this.public.set('proof_did', await this._3id.signJWT(jwt), { noLink: true })
+      }
+    } else {
       // we can just sign an empty JWT as a proof that we own this DID
       await this.public.set('proof_did', await this._3id.signJWT(), { noLink: true })
     }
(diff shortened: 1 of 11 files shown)


src/3box.js
```diff
@@ -286,8 +286,7 @@ class Box extends BoxApi {
    */
   get DID () {
     if (!this._3id) throw new Error('DID: auth required')
-    // TODO - update once verification service supports 3ID
-    return this._3id.muportDID
+    return this._3id.did
```
oed: Pretty sure this needs to be `this._3id.DID`.

src/3id/index.js
```diff
@@ -163,13 +163,6 @@ class ThreeId {
     let docHash = (await this._ipfs.add(Buffer.from(JSON.stringify(doc))))[0].hash
     this._muportDID = 'did:muport:' + docHash
     this.muportFingerprint = utils.sha256Multihash(this.muportDID)
-    const publishToInfura = async () => {
-      const ipfsMini = new IpfsMini(this._muportIpfs)
-      ipfsMini.addJSON(doc, (err, res) => {
-        if (err) console.error(err)
-      })
-    }
-    publishToInfura()
```
oed: :D:D:D

zachferland:
Since proof_did not necessarily need, almost made sense move from that, but requires changes to api to still be used for verification over other proofs. And i guess some people may still get that did from the api there. Also needs both dids available if want to still verify existing proofs without require new ones issued. So made more sense to have simple upgrade path. proof_did still available at same key, just now if there was originally a muport did proof, it will just be wrapped/provided in a new proof signed by 3id. (proof "upgraded" on next write/link) Both dids need to be available so that proofs existing and new can be verified easily. 

Didn't finish yet, and will look at test to make sure they reflect supporting this path.

src/3box.js
oed:
```suggestion
    return this._3id.DID
```
Had a look. It definitely needs to be capitalized.

src/utils/verifier.js
```diff
@@ -32,17 +36,18 @@ module.exports = {
...
    *
-   * @param     {String}            did             The muport DID of the user
+   * @param     {String}            did             The 3ID of the user (or array of equivalent dids)
    * @param     {String}            claim           A did-JWT with claim
    * @return    {Object}                            Object containing username, proof, and the verifier
    */
   verifyTwitter: async (did, claim) => {
+    const dids = typeof did === 'string' ? [did] : did
     if (!claim) return null
     const verified = await didJWT.verifyJWT(claim)
```
oed: 3id-resolver is not registered in this class, so this will only work if you already have an open box. This is kind of annoying.

src/utils/verifier.js
```diff
@@ -88,7 +94,14 @@ module.exports = {
    */
   verifyDID: async (claim) => {
     const verified = await didJWT.verifyJWT(claim)
-    return verified.payload.iss
+    const muport = verified.payload.muport
+    const res = {}
+    if (muport) {
+      const muportDID = await didJWT.verifyJWT(muport).payload.iss
```
oed: Same thing here. 3id-resolver will not be registered if just calling `Box.getVerifiedAccounts`.

src/3box.js
```diff
@@ -417,7 +417,16 @@ class Box extends BoxApi {
...
+    const proofdid = await this.public.get('proof_did')
+
+    if (proofdid) {
+      // if prior muport, re publish with 3id including muport
+      const issuer = didJWT.decodeJWT(proofdid).payload.iss
+      if (issuer.includes('muport')) {
+        const jwt = { muport: proofdid }
+        await this.public.set('proof_did', await this._3id.signJWT(jwt), { noLink: true })
+      }
+    } else {
```
oed: I think this makes sense :)

zachferland:
For resolve 3ids in static api, yeah i think makes sense for time being to resolve through an added endpoint in our api, being the most straight forward. In the future would like want to to resolve through ipfs (or something more standard like a gateway) when there are more people running other nodes. Background or long running instance are also more likely to be available then client side. cc @oed  (also forgot if you mentioned some other options)

oed:
@zachferland No, I think that makes sense for now 👍

zachferland:
requires https://github.com/3box/3box-api/pull/17

zachferland:
Have been seeing intermittent test failures in identity wallet test, here as test failures, and locally sometimes as lock issues. Also saw same on other branches before.

oed:
@zachferland yeah, I've seen those as well for a while. Haven't taken the time to look into them yet!

zachferland:
Will hold another day, will merge once api released with new endpoint

PR #691, opened 2020-01-07
fix(ghost): fixed bug when opening ghost thread while not authenticated

oed:
This fixes a bug where ghost threads don't work when you're not autenticated.

From discord:

 
Kenzo - 3BoxToday at 12:57 AM
@oed 3⃣ 
Now Im able to openThread which is really cool but there seem to be a few issues. The returned thread from openThread should be the same as the thread that was returned the original way via joinThread correct?

Im doing 
    const box = await Box.create(ethereum);
    const thread = await box.openThread(spaceName, threadName, { ghost: true });

    this.setState({ thread, threadJoined: true, }, async () => {
      await this.updateComments();
      await this.updateMembersOnline();

      thread.onUpdate(() => this.updateComments());
      thread.onNewCapabilities(() => this.updateMembersOnline());
    });
 
which is exactly how I was fetching comments the original way, though now it doesnt return any comments.

The original way would fetch comments when onUpdate (so on the second time calling updateComments) - in the new way, the thread does not update (onUpdate doesnt trigger)

On top of this, getting a weird error saying Can not send message if not authenticated, even though there was no attempt to send a message

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -2,13 +2,19 @@ const syncComplete = (res) => {
   console.log('Sync Complete')
   updateProfileData(window.box)
 }
-bopen.addEventListener('click', event => {
+
+Box.create(window.ethereum).then(box => {
+  window.box = box
+  bauth.disabled = false
+  openThread.disabled = false
+})
+
+bauth.addEventListener('click', event => {
 
   window.ethereum.enable().then(addresses => {
-    window.Box.openBox(addresses[0],  window.ethereum, {}).then(box => {
+    window.box.auth([], { address: addresses[0] }).then(() => {
       box.onSyncDone(syncComplete)
-      window.box = box
-      console.log(box)
+      console.log('authed')
 
       controlls.style.display = 'block'
       updateProfileData(box)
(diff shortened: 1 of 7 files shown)


src/ghost.js
```diff
@@ -154,10 +175,10 @@ class GhostThread extends EventEmitter {
    *
    * @param     {Object}    message                 The message
    */
-  async _broadcast (message) {
-    if (!this._3id) throw new Error('Can not send message if not authenticated')
-    const jwt = await this._3id.signJWT(message)
-    this._room.broadcast(jwt)
+  async _broadcast (message, noSignature) {
```
zachferland: looks like for broadcast, noSignature opt no used anywhere?

example/index.js
```diff
@@ -2,13 +2,19 @@ const syncComplete = (res) => {
...
+  window.box = box
+  bauth.disabled = false
+  openThread.disabled = false
+})
+
+bauth.addEventListener('click', event => {
 
   window.ethereum.enable().then(addresses => {
-    window.Box.openBox(addresses[0],  window.ethereum, {}).then(box => {
+    window.box.auth([], { address: addresses[0] }).then(() => {
```
zachferland: 👍

PR #698, opened 2020-01-12
feat: up orbit 0.23, ipfs 0.40

zachferland:
~~Requires  https://github.com/orbitdb/ipfs-log/pull/292  (or another fix) (fix not pulled in to pr)~~

~~TODO~~ 
~~Minimally tested~~


src/3box.js
```diff
@@ -550,20 +550,15 @@ async function initIPFS (ipfs, iframeStore, ipfsOptions) {
...
-      ipfs.on('ready', () => {
-        resolve(ipfs)
-        if (ipfsRepo && typeof window !== 'undefined' && window.indexedDB) {
-          // deletes once db is closed again
-          window.indexedDB.deleteDatabase(ipfsRepo.rootPath)
-        }
-      })
-    })
+
+    ipfs = await IPFS.create(ipfsOptions)
```
oed: Neat!

zachferland:
yeah have to manually run GC right now 

hmm working well so far in usage, but tests are a complete failure

oed:
> yeah have to manually run GC right now

Great! We have to keep an eye out for if that where to change in future versions!

zachferland:
Seems like errors in the tests are going to be blocker here

oed:
It's probably the ipfs setup in the testUtils?

zachferland:
nothing there clearly indicates that so far, but going to try more minimal example of test go see, cant tell what it is so far

zachferland:
downgrading orbit-db-identity-provider allows tests to pass, the totally nondescript error with no trace from level was coming from this and and the orbit-db-keystore. From direct use in 3id. Not sure why older version has been used for a while. 

Should still try to upgrade, but not sure yet what change in lib to cause this. No issues in web.

oed:
Ok, if you run 3box in nodejs do you still get the error? I think I sawsomething similar while running this with latest version of 3box and IDW:
```js
const IdentityWallet = require('identity-wallet')
const Box = require('3box')

;(async() => {
  const authSecret = '0x4ccf180e6d9d20eb9e51f00f34a47fb6ba8b2370dbc70370b1a11aad2034776f'

  const getConsent = async () => true
  const idWallet = new IdentityWallet(getConsent, { authSecret })
  const provider = idWallet.get3idProvider()

  const box = await Box.openBox(null, provider)
  console.log('never gets here!')

  const space = await box.openSpace('demo')
  //await space.public.set('foo', 'bar')
  console.log(await space.public.get('foo'))
})()
```

zachferland:
I was getting same error, but yeah id wallet has issues related to same thing, basically the keystore and identity store we are using not directly through orbit could have better managed lifecycles. Not being closed, or redundant.  (but not clear why keystore worked fine before)

zachferland:
Not sure if there is much i can do about the idw tests right now, they were failing before upgrade on dev branch for me

oed:
Do you know what is causing the issue? What's the context of this specific one? I assume the one related to ipfs update was fixed?

zachferland:
yeah fix for update in commit above, and cleans up the stores further, since before it was making multiple identity/keystores. 

But yeah identity wallet example fails silently (in dev too, maybe broken in last upgrade). This is worth fixing, but may not be worth trying to get those tests run, ideally we dont have unit tests dependent on other tests. Look at circleci, if you have seen that issue about ipfs being offline. But otherwise no info on it.

package.json
```diff
@@ -43,26 +43,31 @@
...
   "dependencies": {
-    "3box-orbitdb-plugins": "^1.0.5",
+    "3box-orbitdb-plugins": "3box/3box-orbitdb-plugins#feat/orbit-23",
     "3id-blockchain-utils": "^0.3.2",
     "3id-resolver": "^0.0.5",
     "@babel/runtime": "^7.4.5",
     "@ethersproject/hdnode": "^5.0.0-beta.133",
     "@ethersproject/wallet": "^5.0.0-beta.133",
     "did-jwt": "^0.2.0",
+    "eth-crypto": "^1.5.1",
```
oed: This shouldn't be here.

zachferland: 👍

src/__tests__/3box.test.js
```diff
@@ -247,7 +239,7 @@ describe('3Box', () => {
 
   afterAll(async () => {
     await testUtils.stopIPFS(ipfs, 0)
-    await testUtils.stopIPFS(ipfsBox, 1)
+    return testUtils.stopIPFS(ipfsBox, 1)
```
oed: I don't get why using await wouldn't work. I thought async functions are converted to promises that resolve when the last await is done?

package.json
```diff
@@ -43,7 +43,7 @@
   },
   "homepage": "https://github.com/3box/3box-js#readme",
   "dependencies": {
-    "3box-orbitdb-plugins": "^1.0.5",
+    "3box-orbitdb-plugins": "3box/3box-orbitdb-plugins#feat/orbit-23",
```
oed: Should we release this before merging this PR?

package.json
```diff
@@ -53,16 +53,20 @@
     "events": "^3.0.0",
     "graphql-request": "^1.8.2",
     "https-did-resolver": "^0.1.0",
-    "ipfs": "^0.36.4",
+    "ipfs": "^0.40.0",
     "ipfs-did-document": "^1.2.3",
+    "ipfs-log": "^4.5.4",
```
oed: Is ipfs-log needed explicitly now?

src/__tests__/testUtils.js
```diff
@@ -23,11 +20,7 @@ const genConf = id => {
 
 module.exports = {
   initIPFS: async (useAltConf) => {
-    return new Promise((resolve, reject) => {
-      let ipfs = new IPFS(genConf(useAltConf))
-      ipfs.on('error', reject)
-      ipfs.on('ready', () => resolve(ipfs))
-    })
+    return IPFS.create(genConf(useAltConf))
```
oed: 🤏

src/__tests__/3box.test.js
zachferland: yeah, tbh not sure what jest is doing with last line awaits, but in idWallet.integration.js the afterAll function was running before the test finished. This worked, i think i also saw a similar problem in the past. 

It was mostly for idWallet tests to pass, so could also not changes lines in other tests right now. It was clearly happen in idWallet, maybe not the others, although they seemed to pass a little more regularly. Still get enough random failures though

oed: Ok, let's keep it like this. It's not really wrong :)

package.json
zachferland: will change in release PR instead, since few days out

PR #702, opened 2020-01-13
fix: static orbitdb cache path, not based id which changes

zachferland:


src/replicator.js
```diff
@@ -89,7 +92,11 @@ class Replicator {
     this._pubsub = new Pubsub(this.ipfs, (await this.ipfs.id()).id)
     // Identity not used, passes ref to 3ID orbit identity provider
     const identity = await Identities.createIdentity({ id: 'nullid' })
-    this._orbitdb = await OrbitDB.createInstance(this.ipfs, { directory: opts.orbitPath, identity })
+    const cachePath = path.join(opts.orbitPath || './orbitdb' , '/cache')
+    const levelDown = Storage(null, {})
+    const cacheStorage = await levelDown.createStore(cachePath)
+    const cache = new Cache(cacheStorage)
```
oed: Would you mind leaving a comment on what this does? (for later reference)

src/replicator.js
```diff
@@ -1,7 +1,10 @@
+const path = require('path')
 const EventEmitter = require('events')
 const OrbitDB = require('orbit-db')
 const Pubsub = require('orbit-db-pubsub')
 const AccessControllers = require('orbit-db-access-controllers')
+const Storage = require('orbit-db-storage-adapter')
+const Cache = require('orbit-db-cache')
```
oed: Maybe name these `OdbStorage` and `OdbCache` for clarity?

src/replicator.js
zachferland: 👍

src/replicator.js
zachferland: good idea

PR #708, opened 2020-01-23
add did as member of replicator class and always send with pin requests

msterle:


src/3box.js
```diff
@@ -86,7 +86,7 @@ class Box extends BoxApi {
     const address = await this._3id.getAddress()
     const rootstoreAddress = address ? await this._getRootstore(address) : null
     if (rootstoreAddress) {
-      await this.replicator.start(rootstoreAddress, { profile: true })
+      await this.replicator.start(rootstoreAddress, this._3id.DID, { profile: true })
```
oed: This won't work since the 3ID won't be authenticated. We should probably create a method for setting the DID and call it after line 100 where `setIdentity` is being called.

zachferland:
locally rebased on dev, and test consistently passing overall (still one test in threads that fails a once in a while, but not related to this)

oed:
@msterle are tests passing locally for you?

zachferland:
Add commit here to help tests: 

Looks like you actually fixed identity wallet aha (@oed code snippet you wrote before works nows for identity wallet). So this test was basically silently failing before and passing, but now working and failing, should be able to help fix that after. Then `3Box › should openBox correctly with normal auth flow, for existing accounts` consistently fails, maybe the change did effect this one. 

Seem like best to run with node versions > 10.12.0 and < 12

src/__tests__/idWallet.integration.js
```diff
@@ -110,8 +110,19 @@ describe('Integration Test: IdentityWallet', () => {
...
+          box.replicator.rootstore.events.off('write', writeHandler)
+          resolve()
+        }
+        firstWrite = false
+      }
+      box.replicator.rootstore.events.on('write', writeHandler)
+    })
     idWallet.addAuthMethod(AUTH_1)
-    await testUtils.delay(1500)
+    await writePromise
```
msterle: nice!

PR #710, opened 2020-01-27
feat: confidential threads

zachferland:
~~No need to review yet~~

~~Implementation sketch, works with first user, little bit of code need for add member/mode, some refactoring, some error handling/constraints and tests~~

### DIFF
--- a/readme-template.md
+++ b/readme-template.md
@@ -174,13 +174,13 @@ const config = await space.private.get('dapp-config')
 ```
 
 ## Threads API (Messaging)
-### Add message threads to your app
-Threads are a shared datastore that enable decentralized communication between users, by allowing one or more users to post messages in a sequence. This functionality is great for adding commenting, chat, messaging, feed, and stream features to your application. Threads are saved within a space and users that join a thread (with the same name, in the same space, and same moderation configs) will be able to communicate in that thread.
+### Add public and confidential message threads to your app
+Threads are a shared datastore that enable decentralized communication between users, by allowing one or more users to post messages in a sequence. This functionality is great for adding commenting, chat, messaging, feed, and stream features to your application. Threads are saved within a space and users that join a thread (with the same name, space, moderation configs, and access configs) will be able to communicate in that thread.
 
 For the fully detailed spec, view the [documentation](https://github.com/3box/3box/blob/master/3IPs/3ip-2.md).
 
-#### Viewing a Thread
-You can get all posts made in a thread without opening a space. This is great for allowing visitors of your site view comments made by other users. This is achieved by calling the `getThread` method on the Box object. A thread can be referenced by all its configuration options or by its address.
+#### Viewing a Public Thread
+You can get all posts made in a public thread without opening a space. This is great for allowing visitors of your site view comments made by other users. This is achieved by calling the `getThread` method on the Box object. A thread can be referenced by all its configuration options or by its address.
 ```js
 const posts = await Box.getThread(spaceName, threadName, firstModerator, membersThread)
 console.log(posts)
@@ -191,12 +191,13 @@ Threads can also be viewed without opening space, or authenticating by calling t
 const posts = await Box.getThreadByAddress(threadAddress)
 console.log(posts)
 ```
-However if applications want to add interactivity to the thread, such as allowing the user to post in a thread or follow updates in a thread, you will need to open their space to enable additional functionality.
+However if applications want to add interactivity to the thread, such as allowing the user to post in a thread or follow updates in a thread, you will need to open their space to enable additional functionality. Same is true for a confidential thread, which requires you autheticate to get access to view the posts in a confidential thread.
 
 #### Interacting with a Thread
 
-##### 1. Joining a thread
-To post in a thread, a user must first join the thread. This will implicitly use the moderation options where the current user is the `firstModerator` and `members` is false.
+##### 1. Creating and joining a thread
+To post in a thread, a user must first join the thread. This will implicitly use the moderation options where the current user is the `firstModerator` and `members` is false. If you are the first to join a public thread it will be created at that point.
+
 ```js
 const thread = await space.joinThread('myThread')
 ```
@@ -207,10 +208,23 @@ A thread can also be given the moderation options when joining. You can pass `fi
 const thread = await space.joinThread('myThread', { firstModerator: 'some3ID', members: true })
 ```
 
-Lastly a thread can be joined by its address.
+To create and join a confidential thread. At creation you will likely want to add other members so that they can read and write messages to the thread.
+
+```js
+const thread = await space.createConfidentialThread('myConfThread')
+```
+
+Lastly an existing thread can be joined by its address. Confidential threads are best referenced by their address. While public threads can be joined by address or known configs as shown above.
 
 ```js
 const thread = await space.joinThreadByAddress('/orbitdb/zdpuAp5QpBKR4BBVTvqe3KXVcNgo4z8Rkp9C5eK38iuEZj3jq/3box.thread.testSpace.testThread')
+```
+
+An address of a thread can be found as follows once joined.
+
+
+```js
+const threadAddress = thread.address
 ```
 
 ##### 2. Posting to a thread
--- a/readme-template.md
+++ b/readme-template.md
@@ -195,8 +195,9 @@ However if applications want to add interactivity to the thread, such as allowin
 
 #### Interacting with a Thread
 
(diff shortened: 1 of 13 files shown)


src/3id/index.js
```diff
@@ -27,6 +33,38 @@ class ThreeId {
...
+  // TODO same as above, move to utils or somwhere
+  async symDecrypt (symKey, message) {
+    const ciphertext = nacl.util.decodeBase64(message.ciphertext)
+    const nonce = nacl.util.decodeBase64(message.nonce)
+    const cleartext = nacl.secretbox.open(ciphertext, nonce, symKey)
+    if (!cleartext) return null
+    const paddedMsg = nacl.util.encodeUTF8(cleartext)
+    return utils.unpad(paddedMsg)
+  }
+
```
oed: Don't these functions exist as helpers in the keyring class?

zachferland:
yes this is true, but on the flipside we cant entirely create a deterministic mapping from those configs to some thread without that included. For example if encrypted in the ac log, there is the possibility that the "same" thread is created in two different places, and without knowing about each other, will try to write that symkey as first entry, later when they do know about each other we end up with two keys for a thread. There is the current restraint that only the firstmod can also create and add the key, so minimized a bit (but even then in a more open network, you have to wait to sync and cant be sure it doesnt exist on other node). Less of issue in current system, soon to more likely be an issue, adding some key id is a lot safer overall. Not sure if much is lost from not being able to derive from config, when you can just ref thread by address, especially as they increase in complexity. I assume by address will best in most cases.

src/3id/index.js
zachferland: yeah, just not exported, so need to be in some shared utils, since sym enc interface in keyring is only for sym key part of keyring

zachferland:
Implementation will look roughly as follows now, number of non happy path conditions to test (or add constraints) 

Most test conditions will be covered by running existing threads tests also over confidential tests. (prob refactor tests a bit) Then some added conditions tested specific to confidential threads. 

Option to be added to example

zachferland:
Remaining 
[] tests for opening by address
[] the final interface for threads in spaces, might change from pass through arg of true or string
[] minor clean up on some encoding/decoding 
[] readme docs

zachferland:
also initialization code not ideal, since requires ac entry to load before getting access to read/write. right now will try to read on load, then will wait for ac updates for 10 secs, if no entry/key by then a no access error is thrown

zachferland:
It appears that opening a thread solely by address has not been working for threads, not clear if this was always the case or was changed over time (seems like former). Will take a bit of work to get that working. Alternatively can be opened by configs still as {name, firstModerator, encKeyId}, and this is stored in space (like current threads) as key:val,  address:{name, firstModerator, encKeyId}

oed:
We need to make it work when joining by addresses. Having to pass "encKeyId" is a horrible dev UX. 
What is that? Why is it needed? <- Not questions that devs should need asking. They shouldn't need to know about the internals of our implementation.

zachferland:
yeah just a status update, not saying address should be left not working and also saying address will not be only way to open still, and the config is as easy/difficult as using the address. 

like said before, either way (even with just first mod) you will have to look up or store a some kind of reference to the thread somewhere. So either way the experience is similar whether that reference is an orbit address, a {name, firstModerator, encKeyId} object or to remove more surface level concerns about encKeyId, you could just rebrand as "thread id" as a string 'name/firstModerator/encKeyid', all those options convey the same info and will referenced/consumed with similar effort

oed:
I just think that "Thread Address" is the thing that's the simplest thing to convey.

zachferland:
@oed can be reviewed

src/3id/index.js
```diff
@@ -10,6 +10,12 @@ Identities.addIdentityProvider(OdbIdentityProvider)
 const utils = require('../utils/index')
 const Keyring = require('./keyring')
 const config = require('../config.js')
+const nacl = require('tweetnacl')
+
+// Redundant key ring
```
oed: "Redundant to key ring"?

src/3id/index.js
```diff
@@ -233,7 +239,7 @@ class ThreeId {
       return utils.callRpc(this._provider, '3id_encrypt', { message, space, to })
     } else {
       const keyring = this._keyringBySpace(space)
-      let paddedMsg = utils.pad(message)
+      let paddedMsg = typeof message === 'string' ? utils.pad(message) : message
```
oed: This doesn't seem correct. If we have an object we should do something like: `utils.pad(JSON.stringify(message))`. Not sure how an object would be handled here otherwise.

src/3id/index.js
```diff
@@ -10,6 +10,12 @@ Identities.addIdentityProvider(OdbIdentityProvider)
 const utils = require('../utils/index')
 const Keyring = require('./keyring')
 const config = require('../config.js')
+const nacl = require('tweetnacl')
+
+// Redundant key ring
+const randomNonce = () => {
+  return nacl.randomBytes(24)
+}
```
oed: Looks like this is available from utils now?

src/3id/utils.js
```diff
@@ -0,0 +1,35 @@
...
+  const cleartext = nacl.secretbox.open(ciphertext, nonce, symKey)
+  if (toBuffer) {
+    return cleartext ? Buffer.from(cleartext) : null
+  }
+  return cleartext ? nacl.util.encodeUTF8(cleartext) : null
+}
+
+const newSymKey = () => {
+  return nacl.box.keyPair().secretKey
+}
```
oed: Why this over `nacl.randomBytes(32)`?

readme-template.md
```diff
@@ -207,30 +209,54 @@ A thread can also be given the moderation options when joining. You can pass `fi
...
+
+```js
+const thread = await space.createConfidentialThread('myConfThread')
+```
+
+At creation you will likely want to add other members so that they can read and write messages to the thread, as shown below.
+
+##### 2. Joining a Thread
+
+An existing public or confidential thread can be joined by its address. Confidential threads are best referenced by their address.
```
oed: Add something about how you get the address of a thread.

src/thread.js
```diff
@@ -49,7 +67,9 @@ class Thread {
   }
 
   async _getThreadAddress () {
-    await this._initConfigs()
+    // TODO WORKS?
```
oed: Remove comment?

src/thread.js
```diff
@@ -38,6 +53,9 @@ class Thread {
     this._subscribe(this._address, { firstModerator: this._firstModerator, members: this._members, name: this._name })
     this._replicator.ensureConnected(this._address, true)
     const timestamp = Math.floor(new Date().getTime() / 1000) // seconds
+
+    if (this._confidential) message = this._symEncrypt(message)
```
oed: We likely need to stringify the message here. Not sure what happens if you try to encrypt an object.

src/thread.js
```diff
@@ -224,6 +297,30 @@ class Thread {
...
+
+  _symEncrypt (message) {
+    const msg = utils.pad(message)
+    return symEncryptBase(msg, this._symKey)
+  }
+
+  _symDecrypt (payload) {
+    const paddedMsg = symDecryptBase(payload.ciphertext, this._symKey, payload.nonce)
+    return utils.unpad(paddedMsg)
+  }
```
oed: maybe JSON stringify / parse can happen here instead?

src/3id/index.js
zachferland: that is for uint8 arrays/buffer, so assumes string or that here, (considered making separate func, but reused more code here instead) could make more specific than string or no string

oed: I see, the way you use it it to encrypt the buffer of the encKey. Makes sense. I guess the way I had implemented this before wasn't general enough to deal with things other than strings.

readme-template.md
zachferland: thats right below

oed: haha 🙈

src/thread.js
zachferland: yeah good catch, should stringify objects somewhere, can add a test for that as well

oed: yep, see comment below 👍

src/3id/index.js
zachferland: yeah exactly, may split in future, but think makes sense to leave shared code path for now

src/3id/utils.js
zachferland: no reason, yeah both 32 rand bytes,  can switch to this

PR #719, opened 2020-02-03
fix(util): Special signature request for authereum

oed:
Check if Authereum and sign with admin key if so.


oed:
Thanks, we where calling `safeSend` directly before with the fromAddress as an extra param. Don't think it's strictly needed but some wallets might be weird and require it.

PR #745, opened 2020-03-26
feat: add iframe cache instantiation and logic

jshrt:
- [ ] Change the url of `3box-shared-cache` to point to `#master`

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -3,7 +3,7 @@ const syncComplete = (res) => {
   updateProfileData(window.box)
 }
 
-Box.create(window.ethereum, { iframeCache: true }).then(box => {
+Box.create(window.ethereum).then(box => {
   window.box = box
   bauth.disabled = false
   openThread.disabled = false
--- a/example/index.js
+++ b/example/index.js
@@ -3,7 +3,7 @@ const syncComplete = (res) => {
   updateProfileData(window.box)
 }
 
-Box.create(window.ethereum).then(box => {
+Box.create(window.ethereum, { iframeCache: true }).then(box => {
   window.box = box
   bauth.disabled = false
   openThread.disabled = false
--- a/src/3box.js
+++ b/src/3box.js
@@ -516,7 +516,6 @@ class Box extends BoxApi {
 }
 
 function createIframeCache () {
-
   const iframe = document.createElement('iframe')
 
   iframe.src = process.env.CACHE_IFRAME_URL
@@ -525,7 +524,7 @@ function createIframeCache () {
   const iframeLoaded = new Promise((resolve, reject) => {
     iframe.onload = () => { resolve() }
   })
-  
+
   document.body.appendChild(iframe)
 
   return iframeLoaded
@@ -539,29 +538,28 @@ function initIPFSRepo (iframeCache) {
   if (typeof window !== 'undefined' && window.indexedDB) {
     const sessionID = utils.randInt(10000)
     ipfsRootPath = 'ipfs/root/' + sessionID
-   
+
     const levelInstance = new LevelStore(ipfsRootPath)
 
     repoOpts = {
       storageBackends: {
-        root: () => levelInstance,
-      },
+        root: () => levelInstance
+      }
     }
 
     if (iframeCache) {
       const iframe = document.querySelector('iframe')
 
       const ipfsRepoStorageProxy = SharedCache.createIpfsStorageProxy({
-        postMessage: (data, origin) => iframe.contentWindow.postMessage(data, "*")
+        postMessage: (data, origin) => iframe.contentWindow.postMessage(data, '*')
       })
-      
+
       repoOpts.storageBackendOptions = {
         blocks: {
-          db: ipfsRepoStorageProxy,
+          db: ipfsRepoStorageProxy
         }
       }
     }
-
   }
 
   const repo = new IPFSRepo('ipfs', repoOpts)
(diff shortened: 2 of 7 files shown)


src/replicator.js
```diff
@@ -93,15 +95,27 @@ class Replicator {
...
+      const iframe = document.querySelector('iframe')
+      const postMessage = iframe.contentWindow.postMessage.bind(iframe.contentWindow)
+
+      const cacheStorageProxy = await SharedCache.createOrbitStorageProxy(cachePath, { postMessage })
+
+      cache = new OdbCache(cacheStorageProxy)
+    } else {
+      const cacheStorage = await levelDown.createStore(cachePath)
+      cache = new OdbCache(cacheStorage)
+    }
```
oed: Doesn't seem right to have the document query selector in this class. Should be abstracted away to happen only in 3box.js maybe? The `cacheStorageProxy` could be sent as an option to the Replicator class perhaps.

src/3box.js
```diff
@@ -529,16 +515,51 @@ class Box extends BoxApi {
   }
 }
 
-function initIPFSRepo () {
+function createIframeCache () {
+  const iframe = document.createElement('iframe')
+
+  iframe.src = process.env.CACHE_IFRAME_URL
```
oed: The source should be hardcoded in the `config.js` module, with an option to change it.

example/index.js
```diff
@@ -3,7 +3,7 @@ const syncComplete = (res) => {
   updateProfileData(window.box)
 }
 
-Box.create(window.ethereum).then(box => {
+Box.create(window.ethereum, { iframeCache: true }).then(box => {
```
oed: Using this should be default. There should be an option to disable it.

webpack.config.js
```diff
@@ -29,6 +30,11 @@ module.exports = {
       }
     ]
   },
+  plugins: [
+    new webpack.EnvironmentPlugin([
+      'CACHE_IFRAME_URL',
+    ])
+  ],
```
oed: This isn't needed if we set it it `config.js`

src/replicator.js
jshrt: :thinking: I will look into it. Passing the cacheStorageProxy as a dependency is good idea as well.

src/3box.js
jshrt: I did not catch `config.js`, thanks for pointing that out!

webpack.config.js
jshrt: Noted

example/index.js
jshrt: Gotcha

src/3box.js
```diff
@@ -18,6 +18,7 @@ const LevelStore = require('datastore-level')
 const didJWT = require('did-jwt')
 
 const SharedCache = require('3box-shared-cache')
+const { iframe_cache_url } = require('./config') // eslint-disable-line
```
oed: You don't need to import this since it's already imported on line 14 :)

src/3box.js
```diff
@@ -529,16 +522,68 @@ class Box extends BoxApi {
   }
 }
 
-function initIPFSRepo () {
+function createIframeCache () {
```
oed: This function should probably be called on pageload. Otherwise loading the iframe will slow things down.

oed: cc @zachferland for input.

src/config.js
```diff
@@ -10,8 +9,7 @@ module.exports = {
   address_server_url: process.env.ADDRESS_SERVER_URL || 'https://beta.3box.io/address-server',
   pinning_node: process.env.PINNING_NODE || '/dnsaddr/ipfs.3box.io/tcp/443/wss/ipfs/QmZvxEpiVNjmNbEKyQGvFzAY1BwmGuuvdUTmcTstQPhyVC',
   pinning_room: process.env.PINNING_ROOM || '3box-pinning',
-  iframe_store_version: process.env.IFRAME_STORE_VERSION || IFRAME_STORE_VERSION,
-  iframe_store_url: process.env.IFRAME_STORE_URL || `https://iframe.3box.io/${IFRAME_STORE_VERSION}/iframe.html`,
+  iframe_cache_url: process.env.IFRAME_CACHE_URL || 'https://iframe.3box.io/iframe.html',
```
oed: Can we change this to be `https://cache.3box.io`?

webpack.config.js
```diff
@@ -29,6 +30,11 @@ module.exports = {
       }
     ]
   },
+  plugins: [
+    new webpack.EnvironmentPlugin([
+      'IFRAME_CACHE_URL',
+    ])
+  ],
```
oed: Is this really needed?

src/config.js
zachferland: yeah

src/3box.js
zachferland: yeah, if in browser env, should optimistically load on pageload, and then if any configuration options passed later on function call differ, reload or ignore (mostly want to optimize default then can add different configuration opts/paths if people need it) 

think we could look to move most of this code out of 3boxjs as well, in attempt to make changes as minimal as possible here and for all iframe code to be more contained. I recently did took that path for idw iframe, after doing it more like this last time, before release of both we can try to share some of the same patterns. Lib could export iframe cache service, with similar interface as well (create, func to get orbit proxy, and func to get ipfs proxy). have to see if anything makes that difficult though

src/config.js
jshrt: yup

src/3box.js
jshrt: Thanks for the catch

zachferland:
Replaced by - https://github.com/3box/3box-js/pull/782

PR #760, opened 2020-04-14
replace address server create and read did functionality with ceramic

msterle:


src/accountLinks.js
```diff
@@ -0,0 +1,60 @@
+const { createLink, validateLink } = require('3id-blockchain-utils')
+
+function convertToCaip10(address) {
+  let [accountAddress, chainId] = address.split('@')
+  if (!chainId) chainId = 'eip155:1'
```
oed, line 5: We can actually get the chainId (the number) from the provider: https://github.com/ethereum/wiki/wiki/JSON-RPC#net_version

src/accountLinks.js
```diff
@@ -0,0 +1,60 @@
...
+      proof = await createLink(did, address, this.provider)
+    }
+    const docId = await this._getDocId(address)
+    const doc = await this._ceramic.loadDocument(docId)
+    await doc.change(proof)
+    return doc.content
+  }
+
+  async _getDocId(address) {
+    const doc = await this._ceramic.createDocument(null, 'account-link', { owners: [convertToCaip10(address)], onlyGenesis: true })
```
oed, line 47: I think this will be super slow if you don't add the `skipWait` option as well.

src/3box.js
```diff
@@ -374,64 +377,6 @@ class Box extends BoxApi {
     }, [])
   }
 
-  async _writeAddressLink (proof) {
-    const validProof = await validateLink(proof)
-    if (!validProof) {
-      throw new Error('tried to write invalid link proof', proof)
-    }
-    if (await this.isAddressLinked({ address: validProof.address })) return true // address already linked
-    await this._writeRootstoreEntry(Replicator.entryTypes.ADDRESS_LINK, proof)
```
oed, line 383: This line writes the proof to the rootstore. Consider writing the address of the account-link to the rootstore perhaps? Later we will write the account-link docid to the "config tile".

src/3box.js
```diff
@@ -17,9 +17,14 @@ const IPFSRepo = require('ipfs-repo')
 const LevelStore = require('datastore-level')
 const didJWT = require('did-jwt')
 
+const Ceramic = require('@ceramicnetwork/ceramic-core').default
+const AccountLinks = require('./accountLinks')
+const isEqual = require('lodash.isequal')
```
oed, line 22: Don't see this being used anywhere.

src/accountLinks.js
```diff
@@ -0,0 +1,69 @@
...
+      proof = await createLink(did, address, this.provider)
+    }
+    await doc.change(proof)
+    return doc.content
+  }
+
+  async read (address) {
+    const docId = await this._getDocId(address)
+    const doc = await this._ceramic.loadDocument(docId)
+    await new Promise(resolve => setTimeout(resolve, 500))
```
oed, line 26: Why do you have a timeout here? To see if you get an update from the network?
You can use `doc.on('update', fn)` which will be triggered when there is an update that happened to the document.

src/accountLinks.js
```diff
@@ -0,0 +1,69 @@
...
+      owners: [this._convertToCaip10(address)],
+      onlyGenesis: true,
+      skipWait: true
+    })
+    return doc.id
+  }
+
+  _convertToCaip10 (address) {
+    let [accountAddress, chainId] = address.split('@')
+    if (!chainId) chainId = 'eip155:' + ((this.provider && this.provider.networkVersion) || '1')
```
oed, line 64: Don't think `this.provider.networkVersion` is a standard property. You should probably get it with a call to the json-rpc.

src/accountLinks.js
msterle: In this case, what we want is to ensure we have the most recent heads from all peers for this document. I don't think either a timeout or waiting for a single update solves the problem - this was just temporary (I should have added a comment to that effect). Any thoughts on how we want to handle this?

oed: I think `doc.on('update', fn)` should have the desired effect. If not we should figure out why and how to solve it better.

msterle: Turns out that I should have been setting the `onlyGenesis` property to `false` in `loadDocument`, because it defaults to `true` - what's the use case that's designed for?

oed: It doesn't seem right that you would want to set `onlyGenesis` to `false`. This should only be considered when you initially create the document. When we load the document we want onlyGenesis to be true since we don't want to create an anchor record or publish the "head" (our genesis CID for the doc). 

The desired behaviour of loadDocument (without options) should be that it waits for any updates from the network and returns once it get's that update.

I guess however since you are now creating the document before with the `_getDocId` function you might be sidestepping the default behaviour.

src/accountLinks.js
```diff
@@ -0,0 +1,78 @@
...
+      proof = await createLink(did, address, this.provider)
+    }
+    await doc.change(proof)
+    return doc.content
+  }
+
+  async read (address) {
+    const docId = await this._getDocId(address)
+    const doc = await this._ceramic.loadDocument(docId, { onlyGenesis: false })
+    return doc.content
```
oed, line 27: Instead of getting docId from the first method call here you should just return the document. We don't need to load the document if we already created it.

src/accountLinks.js
```diff
@@ -0,0 +1,78 @@
...
+
+  async update (address, did, proof = null) {
+    if (!proof) {
+      if (!this.provider) {
+        throw new Error('Provider must be set to update an account link')
+      }
+      proof = await createLink(did, address, this.provider)
+    }
+    const docId = await this._getDocId(address)
+    const doc = await this._ceramic.loadDocument(docId, { onlyGenesis: false })
```
oed, line 38: same thing here. The document is created in the `_getDocId` method, so we might as well use that version of the document.

src/accountLinks.js
```diff
@@ -0,0 +1,78 @@
...
+  // TODO: https://github.com/3box/3box/issues/1022
+  async getAddresses (did) {
+    throw new Error('Not implemented')
+  }
+
+  async _getDocId (address) {
+    const doc = await this._ceramic.createDocument(null, 'account-link', {
+      owners: [await this._convertToCaip10(address)],
+      onlyGenesis: true,
+      skipWait: true
```
oed, line 57: We can set skipWait to false here which means that we will wait for heads from the network.

src/accountLinks.js
```diff
@@ -0,0 +1,78 @@
...
+    })
+    return doc.id
```
oed, line 59: we can just return the doc itself here.

src/accountLinks.js
```diff
@@ -0,0 +1,78 @@
...
+class AccountLinks {
+  constructor (ceramic, provider) {
+    this._ceramic = ceramic
+    this.provider = provider
+  }
+
+  async create (address, did, proof = null) {
+    const doc = await this._ceramic.createDocument(null, 'account-link', {
+      owners: [await this._convertToCaip10(address)]
+    })
```
oed, line 13: I guess we can just use the `_getDocId` method here as well 👍

src/accountLinks.js
msterle: `update` wasn't working because the event is named 'change' :D The 'change' event does work if the document has had updates applied.

If the document *doesn't* exist, though, (the address hasn't been linked yet), if we wait for a `change` event, it waits forever (or until we update that document) - I think that's why I had used a timeout. We could combine the two (i.e. in a Promises race) but the only advantage that will have is potentially reducing the delay, while still being time-bound, and also suffering from the downside that it will assume a document is loaded on the first `change` event it receives.

src/accountLinks.js
msterle: Replacing `_getDocId` with `_getDoc` and the other above suggestions is to me a step towards the optimizations discussed in having a document registry (either here or in ceramic). I suggest we think this through more and implement in a followup ticket - I think the value here is in implementing ceramic account links with the minimum of complexity as a first step.

oed: I do think the approach I'm suggesting does minimize complexity. Also don't think having a document registry in the ceramic node would change how we would interact with it here.

msterle: I had started with creating a document without loading it, and it didn't receive updates, which is why I had to use `loadDocument`. Looking into it, this is likely because a document is created in ceramic before checking if one exists in the _docMap object (which is the document registry I'm referring to), and it in turn isn't placed in the registry. This means there are multiple copies of documents, and they are receiving different update events. I'm going to see if I changing this behaviour in ceramic-core fixes this (https://github.com/ceramicnetwork/js-ceramic/blob/develop/packages/ceramic-core/src/ceramic.ts#L68)

src/accountLinks.js
```diff
@@ -0,0 +1,77 @@
...
+        throw new Error('Provider must be set to create an account link')
+      }
+      proof = await createLink(did, address, this.provider)
+    }
+    const doc = await this._ceramic.createDocument(null, 'account-link', {
+      owners: [await this._convertToCaip10(address)],
+      onlyGenesis: true
+    })
+    if (doc.state.log.length > 1) {
+      throw new Error('Account link already exists')
```
oed, line 22: If this is the case we should check if it already links to the correct DID.
If yes, do nothing. If no, change it.

PR #764, opened 2020-04-20
merge replicator's start and new methods

msterle:
Merge the replicator's `start` and `new` functions.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -89,19 +89,14 @@ class Box extends BoxApi {
 
   async _load (opts = {}) {
     const address = await this._3id.getAddress()
-    const { rootStoreAddress, did } = address ? await this._getLinkedData(address) : {}
-    if (rootStoreAddress) {
-      await this.replicator.start(rootStoreAddress, did, { profile: true })
-      await this.replicator.rootstoreSyncDone
-      const authData = await this.replicator.getAuthData()
-      await this._3id.authenticate(opts.spaces, { authData })
-    } else {
-      await this._3id.authenticate(opts.spaces)
-      const rootstoreName = this._3id.muportFingerprint + '.root'
-      const key = (await this._3id.getPublicKeys(null, true)).signingKey
-      await this.replicator.new(rootstoreName, key, this._3id.DID)
-      this._publishRootStore(this.replicator.rootstore.address.toString())
-    }
+    const did = await this.accountLinks.read(address)
+    console.log('_load', { address, did })
+
+    await this._3id.authenticate(opts.spaces)
+    const rootstoreName = this._3id.muportFingerprint + '.root'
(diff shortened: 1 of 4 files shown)


src/3box.js
```diff
@@ -89,19 +89,14 @@ class Box extends BoxApi {
...
-      const authData = await this.replicator.getAuthData()
-      await this._3id.authenticate(opts.spaces, { authData })
-    } else {
-      await this._3id.authenticate(opts.spaces)
-      const rootstoreName = this._3id.muportFingerprint + '.root'
-      const key = (await this._3id.getPublicKeys(null, true)).signingKey
-      await this.replicator.new(rootstoreName, key, this._3id.DID)
-      this._publishRootStore(this.replicator.rootstore.address.toString())
-    }
+    const did = await this.accountLinks.read(address)
```
oed, line 92: What if no DID has  been linked to from this address yet? Seems like the flow would break then?

src/replicator.js
```diff
@@ -125,16 +125,34 @@ class Replicator {
...
     await this.rootstore.load()
-    this.rootstoreSyncDone = this.syncDB(this.rootstore)
-    const waitForSync = async () => {
+
+    try {
+      this.rootstoreSyncDone = this.syncDB(this.rootstore, { hasResponseTimeout: 2000 })
       await this.rootstoreSyncDone
+    } catch (err) {
+      // timeout error means there isn't a remote database answering HAS_ENTRIES requests,
+      // resolve without syncing
```
oed, line 149: Do I have it wrong that the pinning-node would resolve with  HAS_ENTRIES = 0 if the db isn't present?

oed: If that's not the case now, it would solve the problem if it did.

msterle: you're right, it does - I think I wasn't seeing it before because of the `pinningRoomFilter` not being set correctly

src/3box.js
```diff
@@ -88,20 +88,11 @@ class Box extends BoxApi {
...
-      const authData = await this.replicator.getAuthData()
-      await this._3id.authenticate(opts.spaces, { authData })
-    } else {
-      await this._3id.authenticate(opts.spaces)
-      const rootstoreName = this._3id.muportFingerprint + '.root'
-      const key = (await this._3id.getPublicKeys(null, true)).signingKey
-      await this.replicator.new(rootstoreName, key, this._3id.DID)
-      this._publishRootStore(this.replicator.rootstore.address.toString())
-    }
+    await this._3id.authenticate(opts.spaces)
```
oed, line 91: How can we authenticate without authData? This would not work for IDW. We need to load toostore and get the auth data in order to send it to the auth function.

msterle: This was skipped for now, since auth data will be added through ceramic tiles in the upcoming sprint.

Issue #766, opened 2020-04-21
Error: Provider returned signature from different account than requested


ottodevs:
**Describe the bug**
After automatically updating `3box` to 1.18 (released 1 hour ago) we started receiving this error message in the browser console:
```js
index.js:123 Uncaught (in promise) Error: Provider returned signature from different account than requested
```

After signing this panel:
![image](https://user-images.githubusercontent.com/5030059/79884272-3b19e380-83f5-11ea-877e-8dd526384dcb.png)

We decided trying to pin the `3box` dependency to 1.17.1 in our `package.json` and it automagically fixed the issue, so I wonder what can it be causing that problem with this new update...

zachferland:
@ottodevs thanks for creating the issue, are you using box.auth? (or box.openBox?) and passing an address to that function that matches the address of the account you are signing with?

ottodevs:
> @ottodevs thanks for creating the issue, are you using box.auth? (or box.openBox?) and passing an address to that function that matches the address of the account you are signing with?

Hi @zachferland thank you for the answer! We are using `Box.openBox`. Regarding your second question, I think so, otherwise wouldn't it be failing also with the previous library version? We are transparently passing the current address selected by the user in the provider, metamask in this case.

zachferland:
@ottodevs was able to replicate, and fixed in release 3box@1.18.1, thanks for reporting so quickly! closing this issue now

amcassetti:
I'm getting this error as well when using a Ledger device via MetaMask. Below is the code be implemented: 

`const box = await Box.create(currentProvider)
        await box.auth(["ren-bridge"], {
            address: accounts[0]
        })
        const space = await box.openSpace("ren-bridge")`

PR #770, opened 2020-04-21
Hotfix/address message recover check

zachferland:


zachferland:
Resolves https://github.com/3box/3box-js/issues/766

PR #782, opened 2020-05-06
Feat/iframe cache module

zachferland:
will be replacing - https://github.com/3box/3box-js/pull/745


zachferland:
replaced by - https://github.com/3box/3box-js/pull/881

PR #789, opened 2020-05-12
Release/v1.19.0

msterle:
# Release Notes

## v1.19.0 - 2020-05-12
* chore: upgrade did-jwt and did resolver libraries


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "3box",
-  "version": "1.18.1",
+  "version": "1.19.0",
```
oed, line 3: Why is this a minor release rather than a patch? Seems like it's not changing any apis or adding any new features?

msterle: @oed I don't feel that strongly about it, but my reasoning was that I've been following the strict definition of semver so far, where patch is only for backwards-compatible bug fixes. Because the previous version of 3box-js relied on global instances of resolvers, it is technically possible that this could even be a breaking change (if someone used this library and with their own resolvers), but I think that's a very remote possibility. Still, making it a minor reduces the likelihood of it being automatically upgraded as a dependency.

Happy to change it to patch version, though.

oed: Fair enough, didn't think about that possibility.

zachferland:
didnt merge/release yet, all set for anyone else, or i can release tmrrw morning

PR #796, opened 2020-05-15
feat: 3idconnect with eth provider

zachferland:


oed:
Is this a draft? Please make it a draft PR or add a short comment so that I don't spend time on going though it before it's ready to be looked at :)

zachferland:
Should be able run locally now by running the example!

example/index.js
```diff
@@ -1,19 +1,28 @@
+import Web3 from "web3"
+import web3Modal from './providers.js'
+
+let web3
```
oed, line 4: Why do you import web3 here? We only need the provider right?

oed: I just tested with removing web3 and it works just fine.

example/index.js
```diff
@@ -1,19 +1,28 @@
...
+
+const getProvider = async () => {
+  const provider = await web3Modal.connect();
```
oed, line 7: It really sucks that you have to do this first in order to get the provider. Now we can't create 3box on pageload again. Should be easy to add a change where you can just a pass the provider in the auth stage! (which I think makes more sense anyway).

zachferland:
yeah this is true during the migration, parent window just using 3id from 3boxjs still right now and just sending data into iframe. But after of course only the auth data specifically requested will be created in parent window, not stored and sent to iframe.

example/index.js
zachferland: yeah this was true before for everyone, but yeah agree, wasnt going to do for this release, but easy enough to add as option anyways

example/index.js
zachferland: wasnt sure if all the providers alone had the same interface for getting accounts, but yeah may confusing anyone reading 👍  although example not most friendly code example already

oed:
Ah ok, I see.

zachferland:
same fix applied here as - https://github.com/3box/3box-orbitdb-plugins/pull/25

but the verifying for jwt here is in 3id resolver

PR #805, opened 2020-05-26
feat(3box): add option for specifying Ghost Pinbot address

simonovic86:
- add option for specifying Ghost Pinbot address


src/3box.js
```diff
@@ -107,6 +107,7 @@ class Box extends BoxApi {
    * @param     {String}            opts.pinningNode        A string with an ipfs multi-address to a 3box pinning node
    * @param     {Object}            opts.ipfs               A js-ipfs ipfs object
    * @param     {String}            opts.addressServer      URL of the Address Server
+   * @param     {String}            opts.ghostPinbotAddress URL of the Ghost Pinbot Server
```
oed, line 110: Rename this to `opts.ghostPinbot`. The description should be something like: `MultiAddress of a Ghost Pinbot node`. Since it's not a normal URL :)

simonovic86: renamed :)

PR #835, opened 2020-06-12
Feat/browser feat detection

zachferland:


oed:
Looks good, I assume the reason you include a file is that it's a minimal build of modernizr?

zachferland:
yup you create custom builds only the the feature detections you need, can define same in webpack or configi file but then everyone wont have same build configs

PR #840, opened 2020-06-15
release v1.20.0

zachferland:
Released at 3box@1.20.0-beta.0


RELEASE-NOTES.md
```diff
@@ -1,5 +1,23 @@
...
+const box = await Box.create()
+// Later authenticate user
+const spaces = ['myDapp']
+await box.auth(spaces, { address: '0x12345abcde', provider: ethProvider })
+```
+
+* feat: default to 3ID-Connect, pass an eth provider and 3id-connect will be created in background
+* feat: supported function for browser feature support detection
+* ref: pass a provider at box.auth instead of Box.create, so create can be called on page load.
+* chore: upgrade to ipfs 0.44.0, libp3p-webrtc
```
oed, line 17: `libp3p` is that the new version? 😂

zachferland: hah, fixed that

zachferland:
Went through a round of tests again, looks good to release to me

PR #841, opened 2020-06-15
chore: up ipfs 0.46.0

zachferland:


zachferland:
Also sets explicit libp2p-pubsub@0.4.6

PR #842, opened 2020-06-17
fix: use same level db path as before upgrade

zachferland:
older versions or older version of level prepended level-js- to given path


zachferland:
pushing up, but need test / see if any compat issues, shouldnt be but looks like format changed a bit 

also doesn't have to be released, but would likely decrease infra load and sync times for most if havent upgraded already

zachferland:
not releasing, closing

PR #859, opened 2020-07-10
fix: timeout links which dont resolve ipfs.dag.get

zachferland:
Fix allowed one my oldest accounts to work again (from Oct 2018). One of the address-links in rootstore had a cid which was not available anywhere (ie our pinning node). This may have occurred from number of things, from switching to storing links in rootstore/ipfs, to not pinning links at one point on pinning node (i think this was bug at one point...), etc I did have a second existing link that was valid and resolved (but the other blocked this before fix). This also blocked creating new link, if an old one was not able to resolve.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "3box",
-  "version": "1.20.1",
+  "version": "1.20.2",
   "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


zachferland:
going to increase delay and look at test

PR #862, opened 2020-07-14
fix: Use port number indicated in the documentation

drwasho:
Port number for the example app (indicated here: https://docs.3box.io/try/example-app) uses `3000`, but in the code the port number is`30000`. Either the code or the documentation will need to be changed; if it's the latter, feel free to close the PR.


oed:
@zachferland Can you please update the documentation?

zachferland:
Thanks @drwasho! updated docs, closing here

PR #874, opened 2020-07-29
fix: better error message on openSpace

oed:
Better output of error message in the `openSpace` method.

### DIFF
--- a/src/3box.js
+++ b/src/3box.js
@@ -246,7 +246,7 @@ class Box extends BoxApi {
         if (e.message.includes('User denied message signature.')) {
           throw new Error('User denied space consent.')
         } else {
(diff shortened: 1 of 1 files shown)


src/3box.js
```diff
@@ -246,7 +246,7 @@ class Box extends BoxApi {
         if (e.message.includes('User denied message signature.')) {
           throw new Error('User denied space consent.')
         } else {
-          throw new Error('An error occured while opening space: ', e.message)
+          throw new Error('An error occured while opening space: ', e)
```
dkent600, line 249: should be "occurred"

oed, line 249:
```suggestion
          throw new Error('An error occurred while opening space: ', e)
```

PR #875, opened 2020-07-30
Feat/iframe cache up ipfs

zachferland:


zachferland:
replaced by - https://github.com/3box/3box-js/pull/881

PR #923, opened 2020-10-02
fix ghost thread_3id option passed from space

skylar745:
This was preventing properly passing options to ghost thread


oed (approved):
Looks good, thanks!

